PR #9, opened 2016-01-29
changing service to systemctl to make it work for centos 7.2

ighouri:


schwankl:
No - using the service resource is the right way. I am ok to check node.platform and node.platform_version for 7.2 for a temp workaround tho

PR #10, opened 2016-01-29
fixing chef service resource issue for 7+ platforms

ighouri:


kfilipovski:
@ighouri Can you check if this issue is fixed in newer versions of chef? I would rather keep the chef resource and avoid custom logic based on a platform family. Note that recent versions of Ubuntu are also using systemd so this issue is not going to go away and we don't want to end up adding workarounds all the time.

schwankl:
and this doesnt look tested at all - platform_family doesnt have centos, its rhel

schwankl:
this is only an issue on centos/rhel >7.1  (so far)

PR #26, opened 2016-02-05
cassandra 2.2 ; enhance debian c* install to allow new version without recipe change

schwankl:


ranand19:
for redhat the current code configures logback only for 2.1 .

if dist == "2.1"
  template "/opt/cassandra/conf/logback.xml" do
    source "logback.xml.erb"
    owner "root"
    group "root"
    mode 0644
  end  
end

should add support for 2.2x ?

schwankl:
ya good catch raj

PR #27, opened 2016-02-08
code added to attach data disk to azure computes

kowsalyapalaniappan:
code added to attach data disk (size hardcoded to 1023 GB) and create logical volumes in sdc instead of os_disk(sdb). this is needed because sdb will be wiped off on deallocate action which may occur during maintenance.

Mapping the size of the disk space from openstack computes flavors to azure computes is yet to be done.


components/cookbooks/volume/recipes/add.rb
```diff
@@ -543,12 +551,17 @@
...
+    Chef::Log.info("-------------------------")
     Chef::Log.info("Type : = "+type )
-    Chef::Log.info("-------------------------") 
-    
-     if type == "data" 
-        cmd = "mkfs -t #{_fstype} -f #{_device}"
+    Chef::Log.info("-------------------------")
+
+     if type == "data"
+       if node[:platform] == "centos" && (node[:platform_version]).to_i >= 7
```
schwankl: Should use platform_family here too

PR #28, opened 2016-02-08
Remove Diagnostic Cache Cookbooks from Open Source

ravivangara17:


nzhang1220:
diagnostic-cache is a very important piece of Couchbase pack, primarily doing the monitoring jobs. Without diagnostic-cache, Couchbase will run in the dark.

I just revised the latest `diagnostic-cache`: (1) make sure it could dump metrics into Graphite (and we have Graphite pack on OneOps), (2) remove all walmart keyword. (3) make sure the deployment will finish with both `single` and `redundant` mode.

I think it may be still good to add `diagnostic-cache` back? If the above improvement could release your concern to some extends, I would like to make PR for this.

@schwankl @ravivangara17 thoughts?

screen shot 2016-07-10 at 4 07 08 pm
screen shot 2016-08-11 at 12 32 26 am

PR #29, opened 2016-02-09
activemq fix

msumankumar:


components/cookbooks/activemq/recipes/add.rb
```diff
@@ -30,7 +30,7 @@
 Chef::Log.info("ActiveMq config version : #{config_ver}") if !config_ver.empty?
 
 
-tarball = "/activemq/#{version}/apache-activemq-#{version}-bin.tar.gz"
+tarball = "/activemq/apache-activemq/#{version}/apache-activemq-#{version}-bin.tar.gz"
```
schwankl: the previous convention is correct - yours would not work in apache mirrors...examples:
http://apache.cs.utah.edu/activemq/5.11.3/
http://www.us.apache.org/dist/activemq/5.12.3/

PR #30, opened 2016-02-10
Apply security group definitions from DependsOn to compute node.

ssdowd:
Patch to apply security groups from DependsOn to compute nodes for EC2 (along with the default security group).


schwankl:
your changes brought to light old env based secgroup logic for ec2 - removed it in https://github.com/oneops/circuit-oneops-1/pull/32 which also uses the DependsOn Secgroup

kfilipovski:
closing in favor of #32

PR #35, opened 2016-02-11
Fixed oversized Azure Resource Group Name errors

jitendrakrgupta:
Added Exception handling for oversized Azure Resource Group Name errors with automatic new Resource Group Name suggestion

### DIFF
--- a/components/cookbooks/azure/recipes/get_platform_rg_and_as.rb
+++ b/components/cookbooks/azure/recipes/get_platform_rg_and_as.rb
@@ -1,6 +1,8 @@
 # a little recipe that sets the platform, platform-resource-group and platform-availability-set for azure deployments.
 # several other recipes use this
 
+require File.expand_path('../../libraries/regions.rb', __FILE__)
+
 def generate_rg_name(org,assembly,platform,environment,location)
   Chef::Log.info("Resource Group org: #{org}")
   Chef::Log.info("Resource Group assembly: #{assembly}")
@@ -14,7 +16,7 @@ def generate_rg_name(org,assembly,platform,environment,location)
   #Maximum length of Azure Resource Group Name is 90 characters
    if (resource_group_name.length > 90)
      Chef::Log.info("Resource Group Name is more 90 characters long...Will need to trim it down.")
-     resource_group_name = org[0..15] + '-' + assembly[0..15] + '-' + platform[0..15] + '-' + environment[0..15] + '-' + location
+     resource_group_name = org[0..15] + '-' + assembly[0..15] + '-' + node.workorder.box.ciId.to_s + '-' + environment[0..15] + '-' + AzureRegions::RegionName.abbreviate(location)
   end
   Chef::Log.info("New Resource Group Name = #{resource_group_name}")
   Chef::Log.info("New Resource Group Name Length = #{resource_group_name.length}")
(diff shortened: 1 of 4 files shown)


ReneSaenz:
Hello Jitendra. I would adjust the trimming to all other part (i.e. org, platform, etc) except the region/location. I think we need to full length of the location.

jitendrakrgupta:
Hi Rene,
We have updated the code to exclude location from being trimmed down.Please let us know incase this needs any further changes.

kpgriffith:
@jitendrakrgupta You can run into a problem where the resource group name is not unique enough now and that will cause problems.  For a scenario where you have two platforms, both have names longer than 15 chars and the first 15 are the same, and you are deploying to the same location, the resource group name will be the same, so both will be deployed to the same RG.  If 1 of the platforms is disabled and de-provisioned, then your 1 RG is deleted, and now you've lost everything in Azure, but oneops still thinks one of the platforms is running.

Let's abbreviate the location like we did here: https://github.com/oneops/circuit-oneops-1/blob/master/components/cookbooks/azure/libraries/storage_account.rb
This can be refactored for reusability.

I don't think we should restrict platform name at all.  It is max 32 chars.

components/cookbooks/azure/recipes/get_platform_rg_and_as.rb
```diff
@@ -1,5 +1,26 @@
 # a little recipe that sets the platform, platform-resource-group and platform-availability-set for azure deployments.
 # several other recipes use this
+
+def generate_rg_name(org,assembly,platform,environment,location)
+  Chef::Log.info("Resource Group org: #{org}")
+  Chef::Log.info("Resource Group assembly: #{assembly}")
+  Chef::Log.info("Resource Group Platform: #{platform}")
+  Chef::Log.info("Resource Group Environment: #{environment}")
+  Chef::Log.info("Resource Group location: #{location}")
+  resource_group_name = org + '-' + assembly + '-' + platform + '-' + environment + '-' + location
```
schwankl: you should have the ci_id in here somewhere to make it unique

jitendrakrgupta: @kpgriffith - If I understand your point correctly,two resource group with same names(after trim down of the names) cannot be registered because while registering the second name,it will throw an error that the RG already exists.Therefore the scenario of probably disabling/deleting the incorrect RG may not arise.
Yes,probably we might refactor the location recipe which also has ciID in it to make it unique.

kpgriffith:
After a conversation we are using this format for the resource group name:
org(15) + '-' + assembly(15) + '-' + platform ciId + '-' + environment(15) + '-' + location_abbrv(4)

@jitendrakrgupta Please run a test using Azure DNS, LB and TrafficManager to make sure we are good with this.  All these cookbooks use the resource group name.

ReneSaenz:
I will not check the overall limit  resource_group_name.length > 90
I think it will be simpler if we just use this format

resource_group_name = org[0..15] + '-' + assembly[0..15] + '-' + node.workorder.box.ciId.to_s + '-' + environment[0..15] + '-' + AzureRegions::RegionName.abbreviate(location)

This will work even if the name of the different parts (org, assembly, env) are not over 15 chars.
So do not check if resource_group_name.length > 90

PR #37, opened 2016-02-11
activemq fix

msumankumar:


packs/activemq.rb
```diff
@@ -191,7 +191,8 @@
 resource "java",
          :cookbook => "oneops.1.java",
          :design => true,
-         :requires => { "constraint" => "0..1"},
+         :requires => { :constraint => "1..1",
+                        :services => "compute" },
```
schwankl: why does java need compute cloud service?

PR #39, opened 2016-02-11
activemq fix

msumankumar:


schwankl:
@msumankumar there should not be an activemq payload defined in genericmq - lets move that to the activemq pack

PR #40, opened 2016-02-16
Redis

mtunganati:


components/cookbooks/redisio/metadata.rb
```diff
@@ -0,0 +1,71 @@
+# Metadata for the redisio cookbook 
+# Last Edited and Updated by Ravi Vangara
+name             'Redisio'
+maintainer       'Brian Bianco'
+maintainer_email 'brian.bianco@gmail.com'
+license          'Apache 2.0'
+description      'Installs/Configures redis'
+long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
```
schwankl: remove long_description - will not be syncing md files soon

components/cookbooks/redisio/metadata.rb
```diff
@@ -0,0 +1,71 @@
...
+#          :description => 'Source URL',
+#          :required => 'required',
+#          :format => {
+#              :help => 'location of the redis source distribution',
+#              :category => '1.Global',
+#              :order => 2,
+#	       #:filter => {"all" => {"visible" => "version:eq:2.6.16"}}
+#          }
+
+### support freedom of choice
```
schwankl: why not make it even more free and let the user input the version using a textfield instead of selection?  That way a code change is not needed for 3.0.2

packs/redisio.rb
```diff
@@ -0,0 +1,143 @@
+#include_pack "base"
```
schwankl: can rm that comment - base is included by generic_ring

packs/redisio.rb
```diff
@@ -0,0 +1,143 @@
...
+
+name "redisio"
+description "RedisIO"
+type "Platform"
+category "Database NoSQL"
+
+# Overriding the default compute
+resource 'compute',
+         :cookbook => 'compute',
+         :attributes => {'ostype' => 'default-cloud',
```
schwankl: no ostype attr on compute

PR #46, opened 2016-02-19
Have user configurable directories & seed discovery logic

sbonde:
**1.Configurable Directories : User can configure below directories**  

- commitlog_directory
- data_file_directories
- saved_caches_directory
- Binary install directory

_Once the cluster is created, user can’t update the above directories_

**2. Seed Logic**

- First N (default 1) compute IPs from each DC will be included in the seed list, where N is user provided value (default 1)
- If N > #computes in DC, then all computes from each DC  will be included in the seed list
- If user provided seed list will be given the precedence

**3. End point snitch**
Only below options are available

- GossipingPropertyFileSnitch (Default)
- PropertyFileSnitch
- RackInferringSnitch

_Once the cluster is created, user can’t update the end point snitch_


kfilipovski:
@sbonde i think it's going to get a little confusing with some configuration directives being modeled explicitly as attributes and some defined in the 'configuration_directives' hash attribute.  I believe we do support "fixed keys" hash now so if we wanted to list some of the important directives with default values in that hash (like commitlog_directory, data_file_directories etc.) we should do it there instead of dedicated attributes.

sbonde:
@kfilipovski I had the same question when requirement was provided to me and my team architect Bikram Singh wanted me to do that way. I will pull him for further explanation.

bbhullar:
@kfilipovski As Sanjay mentioned earlier we intentionally separated few attributes because 1. there are subsequent steps needs to be done when modifying them, and 2. they cannot be modified after the environment creation.

Another thought was that if we make them as a part of Hash user could delete , alter them after the deployment. 

I see you mention the fixed hash, is it something that user will not be able to delete? Can you please point to the example?

-Bikram

kfilipovski:
@bbhullar I'm not sure I follow. If they are separate attributes how are you preventing the users from modifying them or deleting the value in those separate attributes?  How is this different from the user modifying them or deleting them in the hash?

schwankl:
I think these should be standard configuration_directives.  An example on how to model them (would be a json hash instead of array tho): https://github.com/oneops/circuit-oneops-1/blob/master/packs/cassandra.rb#L177 ... to prevent users from changing them, you need to use the ciBaseAttributes (previous value) ... if the key-value changes then raise the exception identifying the key-value.

kfilipovski:
@schwankl @bbhullar so this *already is* modeled as a hash here: https://github.com/oneops/circuit-oneops-1/blob/master/components/cookbooks/cassandra/metadata.rb#L135-L142.  There is no way to prevent the user from modifying the values right now after deployment if it's has or dedicated attribute.  And the logic to fail the deployment if the change using ciBaseAttributes can be done if it's a hash or a dedicated attribute.  Hence I don't understand the reasoning behind using separate attribute, it plays no role in terms of preventing the user from making a change.  Maybe the intent is to provide better visibility in those directives up-front, but that can be achieved by simply adding those directives in the hash as default values in the metadata.rb or as default values in the pack resource.

schwankl:
I added the configurable then immutable directories using the config_directives: https://github.com/oneops/circuit-oneops-1/pull/58/files ... your previous way used a single value for data_directories, which was misleading - this way supports multiple.  We should move num_tokens and maybe other attrs to use this approach too.  install_dir should be a separate attr tho - its not part of cassandra.yaml.  The endpoint_snitch can be changed with some conditions - ie can go from PropertyFileSnitch to GossipingPropertyFileSnitch

sbonde:
@schwankl- I will take your changes and create a separate PR.

sbonde:
Closing this PR and will create new after taking @schwankl changes.

PR #52, opened 2016-02-22
Zookeeper

msumankumar:


components/cookbooks/zookeeper/metadata.json
```diff
@@ -0,0 +1,365 @@
+{
```
schwankl: whats this metadata.json for?

components/cookbooks/zookeeper/metadata.rb
```diff
@@ -0,0 +1,274 @@
...
+  :required => 'required',
+  :default               => '/usr/lib/zookeeper',
+  :format => {
+    :category => '1.Global',
+    :help => 'Zookeeper Installation Directory ',
+    :order => 2
+  }   
+
+
+# attribute 'group',
```
schwankl: should rm before merge to master

components/cookbooks/zookeeper/metadata.json
msumankumar: That's just a json format copied to the file. Let me remove it.

PR #58, opened 2016-02-23
cassandra configurable then immutable directories

schwankl:


sbonde:
@schwankl - Could you merge this PR so that I can it?

PR #59, opened 2016-02-23
Node JS changes for actions

ravivangara17:


schwankl:
where does the init.d script get installed from?  the package/epel doest come with it

PR #62, opened 2016-02-24
activemq fix

msumankumar:


schwankl:
queue and topic should be in genericmq, but in activemq should be overridden for the payload and monitors

PR #64, opened 2016-02-24
Unit tests for azuredns:zone.rb library

shaffan-chaudhry-confiz:
Unit tests for azuredns:zone.rb library. Library is refactored and passes rubocop and foodcritic static code analysis. Integration tests verified via OneOps.

- Removed 'puts' statements from recipe. We should use Chef::Log instead of puts.
- Renamed Cookbook name in metadata.rb to lower case letters.
- Added constants.rb file in root folder to reuse constants in project.


kpgriffith:
You actually need the 'puts' statements.  Standard out goes back to the inductor and it uses it.  In this case "puts "***FAULT:FATAL=#{msg}"" the msg will be used and displayed in the UI.

And I'm pretty sure the cookbook name needs to be capitalized.  @schwankl can you confirm?

kpgriffith:
First letter of the cookbook name should be uppercase

PR #65, opened 2016-02-24
Unit tests for azuredns::build_entries_list recipe

shaffan-chaudhry-confiz:
Unit tests for azuredns::build_entries_list recipe. Recipe is refactored and passes rubocop and foodcritic static code analysis. Integration tests verified via OneOps.

- Removed 'puts' statements from recipe. We should use Chef::Log instead of puts.
- Renamed Cookbook name in metadata.rb to lower case letters.
- Removed self dependency from metadata.rb in azuredns cookbook. 
- Added constants.rb file in root folder to reuse constants in project


kpgriffith:
The "puts" are needed for returning information to the Inductor.  put them back please.

schwankl:
there should not be a dup in logic like this.  generating the entries should be the same,  set_ should be different.

kpgriffith:
Yes.  I'm looking at it.

kpgriffith:
@shaffan-chaudhry-confiz please close this PR.  azuredns::build_entries_list.rb is being removed.  Thank You.

shaffan-chaudhry-confiz:
Closing PR ...

PR #81, opened 2016-02-27
align chef version

schwankl:


sureshg:
@schwankl needs to change the `ohai` version too

PR #87, opened 2016-03-04
docker engine (1.9.x , 1.10.x) pack support -part I

sureshg:
Initial docker pack to install and configure docker engine ( > 1.9.x)  on computes.


schwankl:
@sureshg lets rm the .gem file - why is that in the cookbook?

sureshg:
@schwankl that's a workaround until the latest gem (`docker-api-1.26.2.gem`) is available in the internal gen repo. I am working with the repo teams to fix this.

schwankl:
lets rm netscaler changes from this pr too

sureshg:
@schwankl ok..removed the netscaler commit. Will create a separate PR for that.

packs/docker.rb
```diff
@@ -1,95 +1,182 @@
...
+                                :metrics => {
+                                    :up => metric(:unit => '%', :description => 'Percent Up'),
+                                },
+                                :thresholds => {
+                                    :dockerEngineDown => threshold('1m', 'avg', 'up', trigger('<=', 98, 1, 1), reset('>', 95, 1, 1), 'unhealthy')
+                                }
+             }
+         }
+
+resource 'user-app',
```
kfilipovski: @sureshg is this component needed for anything?  if not, let's remove it to avoid wasted workorder, they can always add a user with the generic user from base.

packs/docker.rb
```diff
@@ -1,95 +1,182 @@
...
+resource 'compute',
+         :cookbook => 'oneops.1.compute',
+         :design => true,
+         :attributes => {:size => 'M'}
+
+resource 'secgroup',
+         :cookbook => 'oneops.1.secgroup',
+         :design => true,
+         :attributes => {
+             :inbound => '["22 22 tcp 0.0.0.0/0", "80 80 tcp 0.0.0.0/0", "443 443 tcp 0.0.0.0/0","8080 8080 tcp 0.0.0.0/0","8443 8443 tcp 0.0.0.0/0"]'
```
kfilipovski: @sureshg let's not put extra ports here that are not needed... if the user runs a container on 80 or 8080 or 8443 etc they can open the ports, they should not be open by default.

components/cookbooks/docker_engine/recipes/add.rb
```diff
@@ -0,0 +1,131 @@
...
+extend Docker::Util
+Chef::Resource::RubyBlock.send(:include, Docker::Util)
+
+docker_ver = node.docker_engine.version
+docker_rel = node.docker_engine.release
+docker_pkg = node.docker_engine.package
+docker_svc = node.docker_engine.service
+
+# Check the platform
+exit_with_err "Currently #{docker_pkg} #{docker_ver} is supported only on EL7 (RHEL/CentOS) or later." unless is_platform_supported?
```
kfilipovski: @sureshg when we add pack policies next week this condition should be moved to a pack policy

packs/docker.rb
```diff
@@ -1,95 +1,182 @@
...
+             :services => 'compute'
+         }
+
+resource 'docker_engine',
+         :cookbook => 'oneops.1.docker_engine',
+         :design => true,
+         :requires => {:constraint => '1..1',
+                       :services => 'compute,mirror'},
+         :attributes => {
+             :version => '1.9.1',
```
kfilipovski: @sureshg if it works ok with 1.10 then let's go with 1.10 as default since it's already out.

sureshg: @kfilipovski yeah there are no additional work to be done for 1.10. The thing is it has some issues with `CentOS 7.1` (https://github.com/docker/docker/issues/20484) and i am waiting for `CentOS 7.2 ` to be released internally to make it default in the pack.

packs/docker.rb
sureshg: @kfilipovski can i keep port `80` by default as it required for the container component and it's default demo spring-boot app. Will commit those changes later.

components/cookbooks/docker_engine/recipes/add.rb
sureshg: ok..will remove it once it's available.

packs/docker.rb
sureshg: ok..will update the PR.

sureshg:
@kfilipovski Pushed the changes as per the review comments. Please review it.

PR #96, opened 2016-03-08
inductor arg based add

schwankl:


components/cookbooks/inductor/recipes/add.rb
```diff
@@ -29,35 +29,28 @@
...
+cmd += "--authkey #{inductor[:ciAttributes][:authkey]} "
+cmd += "--additional_java_args \"#{inductor[:ciAttributes][:additional_java_args]}\" "
+cmd += "--env_vars \"#{inductor[:ciAttributes][:env_vars]}\" "
+cmd += "--amq_truststore_location #{inductor[:ciAttributes][:amq_truststore_location]} "
 
 # add/update inductor queue
-execute "inductor add < #{inductor_home}/answers/#{inductor_name}" do
+execute cmd do
   cwd inductor_home
+  user "ooadmin"
```
sureshg: Why we need inductor recipe depends on a specific `ooadmin` user?

PR #102, opened 2016-03-10
adding go pack

ighouri:


packs/go.rb
```diff
@@ -0,0 +1,247 @@
...
+:cookbook => "oneops.1.daemon",
+:design => true,
+:requires => {"constraint" => "1..*"},
+:attributes => {
+    "service_name" => "$OO_LOCAL{artifactId}",
+    "pattern" => "",
+    "use_script_status" => true,
+    "control_script_location" => "/etc/init.d/$OO_LOCAL{artifactId}",
+    "control_script_content" => <<-eoa
+    #!/bin/sh
```
schwankl: This should go in the cookbook not pack

packs/go.rb
```diff
@@ -0,0 +1,247 @@
...
+        },
+    },
+}
+
+resource "volume-app",
+:cookbook => "oneops.1.volume",
+:design => true,
+:requires => { "constraint" => "1..1", "services" => "compute" },
+:attributes => {  "mount_point"   => '/app',
+    "size"          => '10G',
```
schwankl: Should be percent based

packs/go.rb
```diff
@@ -0,0 +1,247 @@
...
+        start)
+        clean
+        start
+        ;;
+        stop)
+        stop
+        ;;
+        restart|reload)
+        stop
+        sleep 10
```
schwankl: Why sleep - can it loop every 1 sec and check if down? Error if up after 10s

packs/go.rb
```diff
@@ -0,0 +1,247 @@
...
+
+variable "repository",
+:description => 'Repository Name',
+:value => ''
+
+variable "shaVersion",
+:description => 'Artifact Download Checksum',
+:value => ''
+
+resource "user-app",
```
schwankl: Why not just use user from base?

schwankl:
This entire pack is just adds an artifact - why is it go specific? Seems like a generic Artifact

PR #114, opened 2016-03-17
Seed list discover logic if not provided

sbonde:
1. Generate seed list with length = seed_count.
2. seed list includes seed_count no. of compute IPs sorted as per ciName. 
3. If seed_count > no. of computes per cloud, then include all IPs in the seed list.
4. If provided (in replace case) given IP will be excluded from the seed list.
5. If user provides seed list then no seed discovery


components/cookbooks/cassandra/recipes/add.rb
```diff
@@ -17,7 +17,7 @@
     Chef::Log.error("platform not supported yet")
 end
 
-
+=begin
```
schwankl: lets rm instead of comment-out

components/cookbooks/cassandra/recipes/add.rb
```diff
@@ -46,6 +46,17 @@
      seeds = tmp_seeds
    end
 end
+
+#Discover seeds if not provided
+if seeds.empty?
+  ipaddress = node[:ipaddress]
```
schwankl: if the node has itself as a seed it will not bootstrap

components/cookbooks/cassandra/recipes/add.rb
sbonde: Removed. (Seed discover as per hostnames)

components/cookbooks/cassandra/recipes/add.rb
sbonde: I have tested below tests and it works-
1. New cluster with single node
2. New cluster multiple nodes in single DC
3. New cluster with multiple nodes in two DCs.

If a node has only self IP, then it creates a cluster.
If node has seeds as different IPs, then it gives enough time for handshake with other(seeds) during this time seeds nodes comes UP as parallel deployment.
If node has seeds as own & different IPs, then it gives enough time for handshake with other(seeds) during this time seeds nodes comes UP as parallel deployment.

PR #117, opened 2016-03-19
User can configure log4j-server.properties or logback.xml

sbonde:
1. User have option as hash to provide key:value for properties.
2. Provided key:value will be transformed to log4j-server.properties or replace with node attribute in logback.xml

For ex.
  "log4j.appender.R.File" => "/var/log/cassandra/system.log",
  "log4j.appender.R.maxFileSize" => "10MB",
  "log4j.appender.R.layout.ConversionPattern" => "%-5level [%thread] %date{ISO8601} %F:%L - %msg%n",
  "log4j.appender.stdout.layout.ConversionPattern" => "%-5level %date{HH:mm:ss,SSS} %msg%n",
  "log4j.rootLogger" => "INFO,stdout,R",
  "com.thinkaurelius.thrift" => "ERROR"


gauravlall:
@sbonde  I think we should be able to model the complete log4j properties so the user can overwrite whats required and able to see what the defaults of log4j are being used . Will probably be simple imo.

PR #134, opened 2016-03-23
errors with couchbase cluster when redundant is used

ssdowd:
data.workorder.payLoad.cb not available in couchbase/library/compenents/bucket_component.rb, use realizedAttrs instead to get admin credentials

node.workorder.payLoad.cm not available in cb_cluster/recipes/base.rb, use realizedAttrs for admin credentials


hemantb:
@ssdowd we are not going to merge this PR as the fix is already available through the PR: https://github.com/oneops/circuit-oneops-1/pull/148 , fix will be available soon for the same in the master branch. Thanks for using the pack and pointing out the issues, as well as appreciate all your inputs here

PR #137, opened 2016-03-23
Play pack : fixing status call

mtunganati:


mtunganati:
@gauravlall  fixed the status call for play service

components/cookbooks/playapp/templates/centos/initd.erb
```diff
@@ -47,8 +47,13 @@ stop() {
 }
 
 status() {
-    #TODO
-    echo
+	if [ -f $APPLICATION_PATH/RUNNING_PID ]; then
+	    PIDFile="$APPLICATION_PATH/RUNNING_PID"
```
gauravlall: We should process table to check if the process is really running state ; if one kills the process then status would fail

components/cookbooks/playapp/templates/ubuntu/initd.erb
```diff
@@ -49,8 +49,16 @@ stop() {
 }
 
 status() {
-    #TODO
-    echo
+	if [ -f $APPLICATION_PATH/RUNNING_PID ]; then
```
gauravlall: Do we Need to add exit 1 ?

components/cookbooks/playapp/recipes/status.rb
```diff
@@ -0,0 +1,5 @@
+app_name=node.workorder.ci[:ciAttributes][:app_name]
+cmd = Mixlib::ShellOut.new("sudo /etc/init.d/#{app_name} status")
```
gauravlall: Do we need sudo here ?

mtunganati: no we don't need it .. was testing a different scenario and left it like that .. removed it

PR #145, opened 2016-03-24
Jboss pack Support in OneOps

sidlinux22:


packs/jboss.rb
```diff
@@ -0,0 +1,161 @@
...
+    :constraint => "1..1",
+    :help => "java programming language environment"
+  },
+  :attributes => {
+  :version => '7'
+  }
+
+
+# depends_on
+[ { :from => 'jboss',     :to => 'os' },
```
gauravlall: Please make java depend on OS , check for dependency on tomcat pack.

PR #147, opened 2016-03-25
convert dns_record to lowercase so it doesn't fail on dns check

sleepyorion:
while testing on AWS dns creation would failed whenever dns_record has uppercase.  This would normalize dns_record to all lowercase.


sleepyorion:
This PR has been tested against aws,rackspace,and openstack.

PR #157, opened 2016-03-28
Adding Component Images for Golang

hemantb:


gauravlall:
We should remove the  go png

PR #163, opened 2016-03-30
Changes to azure recipes to use a proxy if available

kpgriffith:
If the proxy is available as a cloud variable, set it in the azure recipes for work orders.
For action orders, will use a system property if defined in the compute service.


schwankl:
please fix branch conflicts

kpgriffith:
@schwankl Branch conflicts are fixed.

PR #164, opened 2016-03-30
gracefully exiting out of the recipe without chef failure for rackspa…

ighouri:
…ce actions that are not currently supported


schwankl:
if its not supported, it should fail, so the user knows ... not pass silently w/ some info level log saying its not doing anything

PR #169, opened 2016-04-04
added compute service to base pack

ReneSaenz:
This change is necessary because hostname component is looking for the cloud location/region.
The compute component has that info.

### DIFF
--- a/packs/base.rb
+++ b/packs/base.rb
@@ -680,7 +680,7 @@ resource "hostname",
   :design => true,
   :requires => {
     :constraint => "0..1",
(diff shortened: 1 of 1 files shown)


schwankl:
the dns cloud service should have that

schwankl:
the other dns cloud services use the cloud_dns_id attr for what you are trying to solve

ReneSaenz:
The cloud_dns_id attribute comes from the same name field of the DNS service form. This is a free text field, which does not guarantee the correct value.

PR #170, opened 2016-04-04
added netscaler gslb site support for azure sites

carlospuga:


components/cookbooks/fqdn/recipes/add.rb
```diff
@@ -64,6 +64,15 @@
 end
 
 Chef::Log.info("Depends on LB is: #{depends_on_lb}")
+if env.has_key?("global_dns") && env["global_dns"] == "true" && depends_on_lb &&
+   !gdns_service.nil? && gdns_service["ciAttributes"]["gslb_authoritative_servers"] != '[]'
+   if provider =~ /infoblox/
```
schwankl: no - this is not right.  use the gdns service type to route to netscaler or previous logic.

carlospuga: There isn't any new code being introduced to fqdn/recipes/add.rb.  The code is being moved from line 105...108 to 70...73 to align with circuit-main-1.  My changes to circuit-main-1 will only apply to azure_lb not fqdn.

schwankl: there is new code... you are adding if provider =~ /infoblox/ then do some netscaler stuff which is wrong

carlospuga: The context of no "new code" pertains to Netscaler.  Regardless of this PR, the fqdn add recipe was already routing to the netscaler code.  There isn't any new "Netscaler" code being introduced to fqdn/recipes/add.rb.   My intent in moving the already existing netscaler code higher up in the file is so that it runs prior to the fqdn code.  However with that said it shouldn't run if azuredns is the service component.  So to your point on "if provider =~ /infoblox/",  it should really be “if provider !~ /azuredns/” else it affects designate.  I'll make that change and recommit.

schwankl: now there is support for more than 1 provider, add condition to check the cloud service and route.  `if provider =~ /infoblox/ then do some netscaler stuff ` u added is wrong

carlospuga: Clearly as per git the netscaler code was already there.  I have replaced if provider =~ /infoblox/ to if provider !~ /azuredns/

schwankl:
please fix branch conflicts

PR #172, opened 2016-04-04
dotnetcli pack for oneops

murali2020kris:
This installs and unpacks dotnetcli tar file for Linux Centos 7.0. There has been requests from teams to work on DOTNETCLI to build their applications in opensource version of DotNET. This pack will allow them to work on it without manually creating dotnetcli instances. 

In the next release, I will update with more configurable features, ability to download their own packs from internet, etc based on user feedback.


components/cookbooks/dotnetcli/metadata.rb
```diff
@@ -0,0 +1,23 @@
...
...line 22 at SQL statement ; SQL []; ERROR: insert or update on table "dj_rfc_ci_attributes" violates foreign key constraint "dj_rfc_ci_attributes_atrid_fk" Detail: Key (attribute_id)=(19813) is not present in table "md_class_attributes". Where: SQL statement "INSERT INTO dj_rfc_ci_attributes( rfc_attr_id, rfc_id, attribute_id, old_attribute_value, new_attribute_value, owner, comments) VALUES (nextval('dj_pk_seq'), p_rfc_id, p_attribute_id, l_old_attr_value, p_new_attr_value, p_owner, p_comments) returning rfc_attr_id" PL/pgSQL function "dj_upsert_rfc_ci_attr" line 22 at SQL statement; nested exception is org.postgresql.util.PSQLException: ERROR: insert or update on table "dj_rfc_ci_attributes" violates foreign key constraint "dj_rfc_ci_attributes_atrid_fk" Detail: Key (attribute_id)=(19813) is not present in table "md_class_attributes". Where: SQL statement "INSERT INTO dj_rfc_ci_attributes( rfc_attr_id, rfc_id, attribute_id, old_attribute_value, new_attribute_value, owner, comments) VALUES (nextval('dj_pk_seq'), p_rfc_id, p_attribute_id, l_old_attr_value, p_new_attr_value, p_owner, p_comments) returning rfc_attr_id" PL/pgSQL function "dj_upsert_rfc_ci_attr" line 22 at SQL statement
```
schwankl: why is this PSQLException comment?

components/cookbooks/dotnetcli/metadata.rb
```diff
@@ -0,0 +1,23 @@
...
+maintainer_email 'support@oneops.com'
+license          '.NET CLI'
+
+
+grouping 'default',
+  :access => "global",
+  :packages => [ 'base', 'mgmt.catalog', 'mgmt.manifest', 'catalog', 'manifest', 'bom' ]
+attribute 'example_file_name',
+  :description => 'Dotnet Installation Folder',
+  :default => 'usr/share/dotnet',
```
schwankl: relative path to where?  missing leading / ?

components/cookbooks/dotnetcli/recipes/add.rb
```diff
@@ -0,0 +1,50 @@
...
+
+Chef::Log.info("DotNet Installation directory [ #{filePath} ]")
+Chef::Log.info("DotNet destination file [ #{destfile} ]")
+execute "createdirectory" do
+  command "mkdir -p #{filePath}"
+end
+
+
+remote_file "#{destfile}" do
+  source "http://repo.wal-mart.com/content/repositories/walmart/Microsoft/Dotnet/CLI/Centos/0.0.5/Centos-0.0.5.tar.gz"
```
schwankl: no wal-mart hard coded things like this.  use mirror cloud service.

schwankl:
this should not be a pack - should only be a component thats optional in base or some genericdotnet pack that dotnet apps would use.  @kfilipovski do you have any comments?  are dotnet apps just an artifact in a customlb pack?

packs/dotnetcli.rb
```diff
@@ -0,0 +1,46 @@
+include_pack "base"
+name "dotnetcli"
+description "DotNetCLI"
+type "Platform"
+category "Worker Application"
+environment "single", {}
+environment "redundant", {}
+resource "dotnetcli",
+  :cookbook => "1.dotnetcli",
```
schwankl: this wont work - would need to be oneops.1.dotnetcli - see other packs, but imo this should not be a pack to begin with - see comments on pr

murali2020kris: It throws error when I use oneops.1.dotnetcli when I try in web.dev : DEBUG: {"code":400,"errorCode":3012,"message":"there is no class with name mgmt.catalog.oneops.oneops.1.Dotnetcli”}.  It seems to automatically add oneops. at the beginning.

components/cookbooks/dotnetcli/metadata.rb
murali2020kris: This is removed in new version.

components/cookbooks/dotnetcli/metadata.rb
murali2020kris: This is corrected in new version

components/cookbooks/dotnetcli/recipes/add.rb
murali2020kris: Added mirrors in new version

packs/dotnetcli.rb
murali2020kris: This is a parallel to Java pack. Dotnetcli has not evolved to a state where it can be used to build a pack like kestrel or nginx yet. It does not work like a full scale Dotnet Framework. The future versions may fit in that category. But the use case for dotnetcli is just parallel to JAVA pack.

murali2020kris:
@schwankl I get " Non-readable POM /app/looper-work/jobs/circuit-oneops-1/workspace/pom.xml: /app/looper-work/jobs/circuit-oneops-1/workspace/pom.xml (No such file or directory)" in my last commit above. Can you suggest how to fix this please?

PR #174, opened 2016-04-05
Azure blob storage recipes

kowsalyapalaniappan:
Recipes for azure blob storage support.


schwankl:
azure blob is an object store ... azure datadisk is block storage
Azure::ARM::Compute::Models::DataDisk vs https://msdn.microsoft.com/en-us/library/dd135733.aspx

kowsalyapalaniappan:
I have written the recipes to attach the data disk which is the block storage and it's kind of azure blob known as azure page blob . In the azure portal the data disk vhd is labelled as page blob . That's why I am used the terminology . Do you recommend modifying the recipes accordingly ?

schwankl:
so microsoft overused blob for object and block ?  can you post the page where you referenced blob to datadisk ?

kowsalyapalaniappan:
in the link mentioned in your very first comment, there are three different types of blobs described, the datadisk in azure is categorized under page blobs. https://msdn.microsoft.com/en-us/library/dd135733.aspx. the recipes i am attaching the disk(page blob) and deleting it.

kowsalyapalaniappan:
i am closing this PR. I will rename the cookbook and create a new PR.

PR #176, opened 2016-04-06
Adding support for Flamegraphs in OneOps

hemantb:


components/cookbooks/flamegraph/metadata.rb
```diff
@@ -0,0 +1,91 @@
+name               'Flamegraph'
+description        'FlameGraph Installation For Java Apps'
+long_description   IO.read(File.join(File.dirname(__FILE__), 'README.md'))
+version            '1.0'
+maintainer         '@WalmartLabs'
+maintainer_email   'hburma1@email.wal-mart.com'
+license            'Copyright Walmart, All rights reserved.'
```
schwankl: license should be apache 2

PR #198, opened 2016-04-21
Etcd (a distributed key value store) component support

sureshg:
Added support for the latest version 2.3.x


components/cookbooks/etcd/libraries/etcd_util.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    # Checks if the given package is available from the OS repo to install.
+    #
+    # @param pkg_name linux package name
+    # @param version package version
+    #
+    # @return true if the  package is available.
+    #
+    def is_pkg_avail?(pkg_name, version)
+      script = "yum list available #{pkg_name}-#{version}"
```
gauravlall: @sureshg suresh could you please add a check for platform supported here ?

components/cookbooks/etcd/metadata.rb
```diff
@@ -0,0 +1,207 @@
...
+          :format => {
+              :important => true,
+              :help => 'Etcd version',
+              :category => '1.Binary',
+              :order => 1
+          }
+
+attribute 'mirror',
+          :description => 'Binary distribution mirror',
+          :default => '',
```
gauravlall: Should we have default value as github repo ?

sureshg: It is - https://github.com/oneops/circuit-oneops-1/pull/198/files#diff-2e4d3d7a6482ced94749d2076644d7e9R19 . This is the mirror url for for internal use.

components/cookbooks/etcd/libraries/etcd_util.rb
sureshg: @gauravlall added that extra check.

components/cookbooks/etcd/metadata.rb
gauravlall: cool thanks

PR #212, opened 2016-04-25
bug fix for setting the label on the public ip for azure

kpgriffith:
recipe wouldn't compile due to uninitialized errors.  updated the require and include statements to fix it.


oneops-admin:
Build 68

PR #221, opened 2016-04-26
Fix missing resource error in java util.

sureshg:


sureshg:
function is used here - https://github.com/oneops/circuit-oneops-1/blob/master/components/cookbooks/java/libraries/java_util.rb#L220

PR #238, opened 2016-05-04
Objectstore and swift components

umullan:
New component object-store functionality generic for all cloud providers and swift is specific implementation for openstack based cloud.


kfilipovski:
@schwankl can you please review and merge this one too... it should be the same as the internal PR

PR #241, opened 2016-05-05
add initial_user attribute to openstack compute

nzhang1220:
-  add initial_user attribute to openstack compute
-  a new way to install epel repo


schwankl:
@nzhang1220 need to check for nil / compute_service.has_key? 'initial_user'  ... reverted for now

PR #242, opened 2016-05-05
Azure node operations

ReneSaenz:
Refactored azure recipes to use classes
- reboot_node
- status_node
- powercycle_node

Added supporting methods to libraries/VirtualMachine


.chef/knife.rb
```diff
@@ -1,22 +0,0 @@
-log_level :info
```
schwankl: cannot delete this

ReneSaenz:
Added the knife file again. Please review

PR #243, opened 2016-05-05
Fixed hostname component failing issue

zeeshan-arshad-confiz:
Fixed hostname component failing issue.

### DIFF
--- a/components/cookbooks/fqdn/recipes/build_entries_list.rb
+++ b/components/cookbooks/fqdn/recipes/build_entries_list.rb
@@ -164,8 +164,14 @@ else
   end
   is_hostname_entry = true
   ci = os.first
-  dns_name = (ci[:ciAttributes][:hostname] + customer_domain).downcase
 
+  cloud_name = node[:workorder][:cloud][:ciName]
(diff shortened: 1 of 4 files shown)


components/cookbooks/fqdn/recipes/build_entries_list.rb
```diff
@@ -164,8 +164,14 @@ def get_primary_ips(ci,values,customer_domain)
   end
   is_hostname_entry = true
   ci = os.first
-  dns_name = (ci[:ciAttributes][:hostname] + customer_domain).downcase
 
+  cloud_name = node[:workorder][:cloud][:ciName]
+  provider_service = node[:workorder][:services][:dns][cloud_name][:ciClassName].split(".").last.downcase
+  if provider_service =~ /azuredns/
+    dns_name = (ci[:ciAttributes][:hostname]).downcase
```
schwankl: why is the customer domain removed for azure?

zeeshan-arshad-confiz: In case of Azure, hostname already contains customer domain information, when we append(redundant)customer domain to dns_name, it exceeds Azure max limit for allowed labels.

PR #248, opened 2016-05-06
Changes related to blank username and password  for non encrypted password.

adeetikaushal:


bannama:
@adeetikaushal the users.properties file uses node[:activemq][:adminencpwd], but looks like this wont be set if node.activemq.pwdencyenabled is false. can you please check if users.properties is backward compatible with this change when pwdencyenabled is false?

adeetikaushal:
Thanks @bannama. I have  tested and updated the pull request.

PR #252, opened 2016-05-09
Fix json.gem installation and OpenStack floating ip recognition

adanin:
- json.gem issue:
  Migrated to a new 'gem install' command format.

- OpenStack Floating IP issue:
  Mirantis OpenStack 8.0 (Liberty based) was used. The original code didn't pick up a floating IP as a public IP. That leaded to absence of connectivity from OneOps node to a VM instance.

### DIFF
--- a/components/cookbooks/compute/files/default/install_base.sh
+++ b/components/cookbooks/compute/files/default/install_base.sh
@@ -128,7 +128,7 @@ if [ -e /etc/redhat-release ] ; then
    fi	
 fi
 
-gem install json-1.7.7 --no-ri --no-rdoc
+gem install json --version 1.7.7 --no-ri --no-rdoc
 if [ $? -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


components/cookbooks/compute/files/default/install_base.sh
```diff
@@ -128,7 +128,7 @@ if [ -e /etc/redhat-release ] ; then
    fi	
 fi
 
-gem install json-1.7.7 --no-ri --no-rdoc
+gem install json --version 1.7.7 --no-ri --no-rdoc
```
schwankl: history of why "json-1.7.7" was used: an older version of rubygems was not honoring --version and would install some other json* named gem in the local cache/cwd.  I am ok to try --version again and clean this up.  Even tho the install is idempotent, would be nice to skip if already installed

PR #260, opened 2016-05-10
Procedure

sbonde:
1. Nodetool procedure
2. Nodetool action
 
      -Fail if any of the noetool command already in progress.
      -Only sudu user should be allowed to execute all the nodetool commands
     - Monitor the progress as log while nodetool command is in progress.


components/cookbooks/cassandra/recipes/nodetool.rb
```diff
@@ -0,0 +1,36 @@
...
+	result = parts[0]
+end
+if $? != 0
+   puts "***FAULT:FATAL=Failed to execute the command"
+   e = Exception.new("no backtrace")
+   e.set_backtrace("")
+   raise e         
+end
+sleep 5
+begin
```
schwankl: Can you explain this block some more?  seems like a tight loop of returning last 100 lines of the log until the process stops.

sbonde: For example, when user submit the repair command, user want to monitor the log and hence command was submitted as a process/job. Without job, if the command takes long time, oneops timeout and retries the same command again.

schwankl: look at the code again - 1, it wont log any command output if $ != 0, 2: this while true tail -n 100 is bad.  change line 13 to 36 to use shell_out ... example: http://stackoverflow.com/questions/16309808/how-can-i-put-the-output-of-a-chef-execute-resource-into-a-variable

sbonde:
@schwankl - I changed per comment. Could you please review the same.

PR #261, opened 2016-05-11
adding fqdn and compute relationship in glisters pack. same change fo…

ighouri:
…r mysql pack using lbdb pack

### DIFF
--- a/packs/lbdb.rb
+++ b/packs/lbdb.rb
@@ -3,7 +3,6 @@ include_pack "base"
 name "lbdb"
 description "LB fronted Database"
 type "Platform"
-ignore true
 category "Generic"
 
 
@@ -198,7 +197,7 @@ end
     :relation_name => 'DependsOn',
     :from_resource => from,
     :to_resource   => 'lb',
-    :attributes    => { "propagate_to" => 'to', "flex" => false, "min" => 1, "max" => 1 } 
+    :attributes    => { "propagate_to" => 'both', "flex" => false, "min" => 1, "max" => 1 } 
 end
 
 
--- a/packs/lbdb.rb
+++ b/packs/lbdb.rb
@@ -3,6 +3,7 @@ include_pack "base"
 name "lbdb"
 description "LB fronted Database"
 type "Platform"
+ignore true
 category "Generic"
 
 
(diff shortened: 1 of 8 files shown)


packs/lbdb.rb
```diff
@@ -3,7 +3,6 @@
 name "lbdb"
 description "LB fronted Database"
 type "Platform"
-ignore true
```
schwankl: cannot remove that

packs/lbdb.rb
```diff
@@ -198,7 +197,7 @@
     :relation_name => 'DependsOn',
     :from_resource => from,
     :to_resource   => 'lb',
-    :attributes    => { "propagate_to" => 'to', "flex" => false, "min" => 1, "max" => 1 } 
+    :attributes    => { "propagate_to" => 'both', "flex" => false, "min" => 1, "max" => 1 }
```
schwankl: please explain the reason for "both"

ighouri: @schwankl 
Reason propagate_to change as "both" in fqdn::depends_on::lb relation because:
i) we need to have fqdn update happens whenever we do the compute replace 
ii). In this case, from=fqdn and to=lb. Hence propagation defined as "to" will propagate lb only and propagation defined as "from" will only propagate fqdn only. 
iii). currently propagate_to defined as "to" but since we need propagate_to defined as "from" so making propgate_to as "both" will satisfy both the conditions. 
iv). Please note similar is the case for lb::depends_on::compute relation i.e. defined as "both" in https://github.com/oneops/circuit-oneops-1/blob/master/packs/lbdb.rb#L191

packs/lbdb.rb
ighouri: @schwankl 
I made mysql pack inheriting lbdb pack and assumed lbdb pack will get ignored while uploading to CMS if I put ignore as true. 
Please correct if my assumption is not accurate. Thanks

ignore	Ignore/Delete this pack when loading to CMS. Mainly used for defining abstract base packs, which other packs inherit	Default is false

schwankl: ignore true is needed so it doesnt appear in the ui as an option

ighouri: ahh okay .. removing it now.

PR #271, opened 2016-05-14
exiting out of the recipe is environment is single.. Also removing un…

ighouri:
…ncessary chef logs

### DIFF
--- a/components/cookbooks/glusterfs/libraries/glusterfs.rb
+++ b/components/cookbooks/glusterfs/libraries/glusterfs.rb
@@ -7,12 +7,3 @@ def find_bricks(index,replicas,length)
   end
   return bricks
 end
-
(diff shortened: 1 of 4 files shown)


components/cookbooks/glusterfs/libraries/glusterfs.rb
```diff
@@ -16,3 +16,12 @@ def check_for_error_message(message)
     return "success"
   end
 end
+
+def check_environment_availability(environment)
```
sureshg: Why we need this function? It's returns the same value.

ighouri: No suresh. It returns values (either single or redundant) depending on the environment configuration as you can see the function definition.

PR #276, opened 2016-05-17
Small fix to etcd cookbook

nzhang1220:
while working on the improvement of postgresql cookbook (especially for
easier HA setup) by following the article below,

https://github.com/compose/governor.git

I would like to use `etcd` in postgres cookbook, so I simply add the
following into the postgres cookbook (add.rb),

include_recipe "etcd::add"

however it needs some tweaks to make it finally work with another
cookbook (which is `postgres` in my case), which were incorporated in this commit.

Please feel free to comment if those tweaks are valid or not.

### DIFF
--- a/components/cookbooks/etcd/libraries/etcd_util.rb
+++ b/components/cookbooks/etcd/libraries/etcd_util.rb
@@ -66,24 +66,27 @@ module Etcd
       # Search for cookbook mirror
       base_url = ''
       base_url = mirror[cookbook] if !mirror.nil? && mirror.has_key?(cookbook)
-      version = node[cookbook][:version]
+      version = node.etcd.version
 
       if base_url.empty?
         Chef::Log.info("#{cookbook} mirror service is empty. Checking for any http(s) mirror/official release URL.")
-        base_url = node[cookbook][:mirror]
-        base_url = node[cookbook][:release_url] if base_url.empty?
+        if !node[cookbook][:mirror].nil? && !node[cookbook][:mirror].empty?
+	  base_url = node[cookbook][:mirror]
+	end
+	base_url = node.etcd.release_url if base_url.empty?
       end
+      Chef::Log.info("base_url: #{base_url}")
 
       # Replace any $version/$arch/$extn placeholder variables present in the URL
       # e.x: https://github.com/coreos/etcd/releases/download/v$version/etcd-v$version-$arch.$extn
       base_url = base_url.gsub('$version', version)
-                         .gsub('$arch', node[cookbook][:arch])
-                         .gsub('$extn', node[cookbook][:extn])
+                         .gsub('$arch', node.etcd.arch)
+                         .gsub('$extn', node.etcd.extn)
       exit_with_err("Invalid package base URL: #{base_url}") unless url_valid?(base_url)
 
       file_name = File.basename(URI.parse(base_url).path)
       Chef::Log.info("Package url: #{base_url}, filename: #{file_name}")
-      return base_url, file_name
+      return File.dirname(base_url), file_name
     end
 
     # Checks if the given string is a valid http/https URL
--- a/components/cookbooks/etcd/recipes/add.rb
+++ b/components/cookbooks/etcd/recipes/add.rb
@@ -31,6 +31,11 @@ if !is_pkg_avail?('etcd', version)
     source "#{base_url}/#{file_name}"
   end
 
+  directory extract_path do
+    recursive true
+    action :delete
+  end
+
   [node.etcd.working_location, node.etcd.conf_location, extract_path].each do |dir|
     directory "#{dir}" do
       user 'root'


components/cookbooks/etcd/recipes/add.rb
```diff
@@ -31,6 +31,11 @@
     source "#{base_url}/#{file_name}"
   end
 
+  directory extract_path do
```
nzhang1220: seems that without 'delete' the extract_path, the re-deployment or re-installation of etcd will fail, especially `Untar` command, complaining that extract_path is not empty

components/cookbooks/etcd/libraries/etcd_util.rb
```diff
@@ -66,24 +66,27 @@ def get_pkg_location(cookbook)
...
-                         .gsub('$arch', node[cookbook][:arch])
-                         .gsub('$extn', node[cookbook][:extn])
+                         .gsub('$arch', node.etcd.arch)
+                         .gsub('$extn', node.etcd.extn)
       exit_with_err("Invalid package base URL: #{base_url}") unless url_valid?(base_url)
 
       file_name = File.basename(URI.parse(base_url).path)
       Chef::Log.info("Package url: #{base_url}, filename: #{file_name}")
-      return base_url, file_name
+      return File.dirname(base_url), file_name
```
nzhang1220: Without File.dirname(), base_url looks like `https://github.com/coreos/etcd/releases/download/v2.3.1/etcd-v2.3.1-linux-amd64.tar.gz`. Later when `file_name` appends to it, the final link will look like:

`https://github.com/coreos/etcd/releases/download/v2.3.1/etcd-v2.3.1-linux-amd64.tar.gz/etcd-v2.3.1-linux-amd64.tar.gz`

which has a duplicate of `etcd-v2.3.1-linux-amd64.tar.gz`

components/cookbooks/etcd/libraries/etcd_util.rb
```diff
@@ -66,24 +66,27 @@ def get_pkg_location(cookbook)
       # Search for cookbook mirror
       base_url = ''
       base_url = mirror[cookbook] if !mirror.nil? && mirror.has_key?(cookbook)
-      version = node[cookbook][:version]
+      version = node.etcd.version
```
nzhang1220: If using `etcd` in another cookbook (e.g. postgres), we have to define `node[:postgresql][:version] = 2.3.1`, which is a little bit confusing, because 2.3.1 is for `etcd`, not for `postgresql`

components/cookbooks/etcd/libraries/etcd_util.rb
```diff
@@ -66,24 +66,27 @@ def get_pkg_location(cookbook)
...
       end
+      Chef::Log.info("base_url: #{base_url}")
 
       # Replace any $version/$arch/$extn placeholder variables present in the URL
       # e.x: https://github.com/coreos/etcd/releases/download/v$version/etcd-v$version-$arch.$extn
       base_url = base_url.gsub('$version', version)
-                         .gsub('$arch', node[cookbook][:arch])
-                         .gsub('$extn', node[cookbook][:extn])
+                         .gsub('$arch', node.etcd.arch)
+                         .gsub('$extn', node.etcd.extn)
```
nzhang1220: same reason as above about `node[cookbook][:version]`

components/cookbooks/etcd/libraries/etcd_util.rb
```diff
@@ -66,24 +66,27 @@ def get_pkg_location(cookbook)
...
       base_url = ''
       base_url = mirror[cookbook] if !mirror.nil? && mirror.has_key?(cookbook)
-      version = node[cookbook][:version]
+      version = node.etcd.version
 
       if base_url.empty?
         Chef::Log.info("#{cookbook} mirror service is empty. Checking for any http(s) mirror/official release URL.")
-        base_url = node[cookbook][:mirror]
-        base_url = node[cookbook][:release_url] if base_url.empty?
+        if !node[cookbook][:mirror].nil? && !node[cookbook][:mirror].empty?
```
nzhang1220: not sure if most cookbook will have `mirror` attribute in place. If yes, it may make more sense to define `mirror` attribute for the cookbook (e.g. postgres, where could I download postgres), rather than for `etcd`?

But anyway, if we still prefer `node[cookbook][:mirror]` to be the mirror for `etcd`, then we may want to check if it is `nil` or not.

kfilipovski:
@nzhang1220 what if you just include the etcd component in the pack instead of directly including the cookbook inside the cookbook... This way you get the full lifecycle of etcd. Also, if you need any customizations you can set different default attribute values for etcd in the Postgres pack.

nzhang1220:
@kfilipovski totally make sense to me - this a very good advice.

PR #277, opened 2016-05-17
fixed the download path issue

swathisingamsetty12:


components/cookbooks/zookeeper/recipes/add.rb
```diff
@@ -7,16 +7,16 @@
 #
 
 zk_basename = "zookeeper-#{node[:zookeeper][:version]}"
-zk_download_location="#{node[:zookeeper][:mirror]}/zookeeper-#{node[:zookeeper][:version]}/#{zk_basename}.tar.gz"
-
- Chef::Log.info("download locaction #{zk_download_location} ")
+ci = node.workorder.rfcCi.ciAttributes;
```
schwankl: no need for semicolon

swathisingamsetty12: removed the semicolon

PR #279, opened 2016-05-18
Opscenter

sbonde:
1. Added opscenter as an optional component.
2. User can provide opscenter server host as one of the existing node or outside oneops.
3. If user didn't provide the server host, consider first node as server node.


components/cookbooks/opscenter/libraries/opscenter_util.rb
```diff
@@ -0,0 +1,142 @@
...
+			opscenter_ciName = ''
+			payload = node.workorder.payLoad
+			if !payload.has_key?("Opscenter")
+				puts "Opscenter payload not found, "
+				puts "***FAULT:FATAL=Opscenter payload not found, please pull the design"
+				e = Exception.new("no backtrace")
+				e.set_backtrace("")
+				raise e
+			end
+			opscenter_nodes = payload.Opscenter
```
schwankl: dont understand why you need bom.Opscenter custom payload on a bom.Opscenter workorder ... the opscenter_url should be on node.workorder.rfcCi.ciAttributes.opscenter_url

sbonde: Actually when user doesn't provide opscenter server host, in that case I consider the first node as server(sorted by cloudId & compute ciName) .
In subsequent deployments, I always wants to keep the same node as server but if user add a new cloud having lesser  cloudId will be consider as new server host and hence will have multiple servers.
We don't expect user always to provide the server host.
In order to configure any node as agent, I need server host and hence I need to iterate over existing opscenter nodes to find which one is the server.

If there is any other way I can keep the server always same unless user enters the host, please suggest I can modify.

PR #280, opened 2016-05-18
PostgreSQL cookbook improvement - milestone #1

nzhang1220:
This is the first PR that aims to (1) follow the HA PostgreSQL best practices discussed in the following two articles, and (2) support OneOps primary and secondary clouds. Ultimately, PostgreSQL pack should support seamless failover across primary and secondary clouds.

https://www.compose.io/articles/high-availability-for-postgresql-batteries-not-included/

https://github.com/compose/governor/

**The first PR will achieve the following goal:**
support PostgreSQL deployment in secondary clouds as a pure backup of PostgreSQL. When all primary clouds are down, PostgreSQL may become read-only. Simply flipping primary and secondary clouds is not a seamless failover solution across DC level. The seamless failover solution will be handled in the next few PRs.

**High-level summary of this PR:**
`Etcd`: only installed, configured and run in primary clouds. If some of primary clouds (>50% of nodes in primary clouds) are down, the PostgreSQL may become read-only, since the new master PostgreSQL can not be elected if Etcd service is NOT running.

`PostgreSQL`: deployed both on primary and secondary clouds, the "master" PostgreSQL must be from primary clouds, when the master PostgreSQL is failed, the new master (or "potential master") PostgreSQL must be elected from primary clouds. PostgreSQL from secondary clouds only replicate from master PostgreSQL and has no chance to become master.

`HAproxy`: only installed, configured and run in primary clouds. If all primary clouds are down, the HAproxy service is down.

**What are changed in this PR:**
`Etcd`: add simple "primary - secondary" cloud awareness: if the cloud is secondary, do not deploy. The current Etcd cookbook do not differentiate b/w primary and secondary clouds.

`PostgreSQL`: change Governor code to let the PostgreSQL from secondary clouds will not be considered for new master. In this patch, without creating more dependency, 

`HAproxy`: add simple "primary - secondary" cloud awareness: if the cloud is secondary, do not deploy. add `check_port` attribute in metadata, which will be used in `haproxy.cfg`.

`genericlb.rb`: make the propagation rule b/w fqdn and lb in both direction.


components/cookbooks/etcd/recipes/replace.rb
```diff
@@ -4,5 +4,12 @@
 # Author : OneOps
 # Apache License, Version 2.0
 
-include_recipe 'etcd::delete'
+etcd_exist = `systemctl list-units | grep etcd`
+
+Chef::Log.info("etcd_exist:" + etcd_exist)
+
+if !etcd_exist.empty?
```
nzhang1220: Delete etcd, only if `etcd` is installed and configured

components/cookbooks/haproxy/metadata.rb
```diff
@@ -160,7 +160,15 @@
     :order => 10
   }         
   
-             
+attribute 'check_tcp_port',
```
nzhang1220: postgresql HA solution needs to ping HAProxy for a TCP port (e.g. 15432) to check which postgresql instance is the leader. Therefore, need to explicitly expose an attribute to specify the port number

components/cookbooks/postgresql/recipes/add.rb
```diff
@@ -1,359 +1,162 @@
...
+    include_recipe "postgresql::server_debian"
+end
 
+execute "chown-data-dir" do
+    command "chown -R postgres:postgres #{node['postgresql']['data']}"
+    user "root"
+    action :run
+end
 
+execute "chmod-700-data-dir" do
```
nzhang1220: postgresql will not start, if the data directory is not 700

components/cookbooks/postgresql/recipes/client.rb
```diff
@@ -1,27 +0,0 @@
-#
```
nzhang1220: Originally, it was a ruby file without actually doing anything

components/cookbooks/postgresql/recipes/gen_config.rb
```diff
@@ -1,33 +0,0 @@
-
```
nzhang1220: Governor PostgreSQL HA solution will ingest postgres config runtime (the config is actually saved in its own yml file), rather than original `postgresql.conf`.

Same for `pg_hba.conf`

Move `check_sql_pg.rb` and `pg_stats.yaml` to `add.rb`, so that this file could be deleted

packs/postgresql.rb
```diff
@@ -9,21 +9,27 @@
...
          :attributes => {
-             "inbound" => '[ "22 22 tcp 0.0.0.0/0", "5432 5432 tcp 0.0.0.0/0"]'
+             "inbound" => '[ "22 22 tcp 0.0.0.0/0", "5432 5432 tcp 0.0.0.0/0", "2379 2379 tcp 0.0.0.0/0", "2380 2380 tcp 0.0.0.0/0", "5000 5000 tcp 0.0.0.0/0", "15432 15432 tcp 0.0.0.0/0" ]'
          },
          :requires => {
              :constraint => "1..1",
              :services => "compute"
          }
 
+#resource "database",
```
schwankl: why is database commented out?

packs/postgresql.rb
```diff
@@ -255,41 +304,74 @@
...
+    relation "#{from}::depends_on::hostname",
     :relation_name => 'DependsOn',
     :from_resource => from,
-    :to_resource   => 'database',
-    :attributes    => { "flex" => false, "min" => 1, "max" => 1 } 
+    :to_resource   => 'hostname',
+    :attributes    => { "flex" => false, "min" => 1, "max" => 1 }
 end
 
+#[ 'postgresql' ].each do |from|
```
schwankl: remove obsolete relations

packs/postgresql.rb
nzhang1220: I guess (maybe wrong) Governor initially creates `pg_hba.conf` by only having the `replicator` user to access `replication` database (which does not exist initially). This is what `pg_hba.conf` initially contains:

`host    replication replicator 0.0.0.0/0   trust`

in order to have other users to access the database, what I did was: (1) stop governor, (2) amend another line like blow, (3) start governor. (These three steps is documented in README.md)

`host    all all 0.0.0.0/0   trust`

I think `database` component will automatically create users and database. but sure not if the following will work without adding `host all all 0.0.0.0/ trust`

https://github.com/oneops/circuit-oneops-1/blob/master/components/cookbooks/database/recipes/postgresql.rb#L19-L25

packs/postgresql.rb
nzhang1220: Will do eventually after all concerns and questions are cleared.

packs/postgresql.rb
nzhang1220: or if we still want to run `database` component, here may be a solution (for the first-time deployment):
(1) start governor-based postgres (which will generate `pg_hba.conf`)
(2) add `host all all 0.0.0.0/0 trust` to `pg_hba.conf`
(3) restart governor-based postgres
Then `database` should run without access issue.

PR #294, opened 2016-05-24
gslb status fix for non glb env ; cleaner infoblox fail

schwankl:


components/cookbooks/fqdn/recipes/set_dns_entries_infoblox.rb
```diff
@@ -198,24 +200,23 @@ def get_existing_dns (dns_name,ns)
...
-        delete_dns(dns_name, del_value)
-      end
-
-      resp_obj = node.infoblox_conn.request(
-        :method => :post,
-        :path => "/wapi/v1.0/record:#{dns_type}",
-        :body => JSON.dump(record))
-
-      Chef::Log.info("retry response: #{resp_obj.inspect}")
+    infoblox_resp_obj = resp_obj.inspect
```
sureshg: Shouldn't this be the just `resp_obj` instead of `resp_obj.inspect`?

schwankl: no - the deserialization is tried below.  infoblox can return both string and json encoded response

PR #295, opened 2016-05-24
Fixing Correct way to get nginx status

hemantb:


components/cookbooks/nginx/recipes/status.rb
```diff
@@ -2,6 +2,6 @@
 # Cookbook Name:: nginx
 # Recipe:: status
 
-service "nginx" do
-  action :status
-end
+cmd = Mixlib::ShellOut.new("service nginx status")
```
sureshg: Shouldn't `execute` resource be more portable  that using shellout directly? , e:g: https://github.com/oneops/circuit-oneops-1/blob/master/components/cookbooks/docker_engine/recipes/status.rb#L8-L11

hemantb:
@sureshg  I did that initially but the issue was that when the process was stopped the recipe would fail,
Ran systemctl status nginx returned 3
---- End output of systemctl status nginx ----
STDERR:
[2016-05-25T04:14:11+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)

But anycase have done ignore_failure true so this should be fine

components/cookbooks/nginx/recipes/status.rb
```diff
@@ -2,6 +2,8 @@
 # Cookbook Name:: nginx
 # Recipe:: status
 
-service "nginx" do
-  action :status
+execute 'systemctl status nginx' do
+  user 'root'
```
gauravlall: @hemantb Do we really need this to be run as root ?  Th daemon based monitoring for nagios would fail .

PR #332, opened 2016-06-16
AWS reboot, powercycle, replace bug fix

ravivangara17:


schwankl:
ec2 uses the generic fog recipe - dont create a del_node_ec2

PR #333, opened 2016-06-16
Get gslb host from gdns service instead of lb service for azure clouds

kowsalyapalaniappan:
Get gslb host from gdns service instead of lb service for azure clouds.

### DIFF
--- a/components/cookbooks/netscaler/recipes/get_gslb_connection.rb
+++ b/components/cookbooks/netscaler/recipes/get_gslb_connection.rb
@@ -12,7 +12,7 @@ if node.has_key?("ns_conn")
 end
 
 cloud_name = node[:workorder][:cloud][:ciName]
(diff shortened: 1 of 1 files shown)


schwankl:
use the class instead of name ... someone could name an azure cloud "dev" and this would not work.   This also mandates that every lb have gdns ; doesnt sound good.  This just to config the current misconfig of gdns services for azure?

kowsalyapalaniappan:
ok i will use the class name instead of cloud name. i believe this recipe is called only when gdns is enabled in the environment . if gdns is enabled in environment, gdns service has to be already added to the cloud , otherwise the deployment plan will not be generated. yes this is a temporary solution until we have virtual global lb service running on azure datacenters.

kowsalyapalaniappan:
Used location field instead of cloud name to identify the azure cloud.

PR #350, opened 2016-07-01
fix nodejs to use binary only: public repo or internal nexus proxy of

schwankl:
public repo


sleepyorion:
Changes are minor, overall it looks good.

PR #363, opened 2016-07-21
Etcd secondary cloud

nzhang1220:
Assume that the nodes from secondary clouds will form a new Etcd
cluster, which is independent of the Etcd cluster in primary clouds.

The goal of the following method is to identify which IPs belong to secondary clouds.
It assumes that nodes in secondary clouds could access the primary Etcd by CURL.
For now, since FQDN always maps to the nodes in primary clouds and nodes do not
communicate with each other directly about who is from primary clouds, we could let
nodes in secondary clouds to take advantage of FQDN to ask who are the members of
primary Etcd cluster by a CURL (http GET call).

For example:

curl -L http://fqdn_url:2379/v2/members

The response is JSON object which contains the Etcd members in primary
 Etcd cluster:

{"members":[{"id":"14ae5702cb4236f9","name":"compute-238213-2","peerURLs":
 ["http://10.65.227.132:2380"],"clientURLs":["http://10.65.227.132:2379"]},
 {"id":"656b212b423387e4","name":"compute-238213-1","peerURLs":["http://10.65.226.26:2380"],
 "clientURLs":["http://10.65.226.26:2379"]}]}

By removing the above returned IPs from workorder.PayLoad.computes, the remaining IPs should
belong to secondary clouds.

Note: there may be some assumptions for above approach:
1. Etcd component has to depend on "hostname" component with PTR enabled.
2. Etcd in primary clouds is in working state
3. all Etcd nodes in primary clouds are functioning correctly


oneops-admin:
Build 225

oneops-admin:
Build 226

oneops-admin:
Build 227

oneops-admin:
Build 228

oneops-admin:
Build 229

oneops-admin:
Build 230

oneops-admin:
Build 231

oneops-admin:
Build 232

PR #370, opened 2016-07-26
lbdb: bump up default number of compute

nzhang1220:
Postgres Governor depends on lbdb and uses Etcd for the leader election,
where Etcd is best run on the minimum of 3 nodes, which leads to better experiences for the first-time users.


schwankl:
lets override the relation in the postgresql-governor pack level

nzhang1220:
@schwankl which of the following will work?

[ 'lb' ].each do |from|
  relation "#{from}::depends_on::compute",
  :only => [ 'redundant' ],
  :from_resource => from,
  :to_resource   => 'compute',
  :attributes    => { "propagate_to" => 'both', "flex" => true, "current" =>3, "min" => 1, "max" => 3}
end

OR

[ 'lb' ].each do |from|
  relation "#{from}::depends_on::compute",
    :except => [ '_default', 'single' ],
    :design => true,
    :from_resource => from,
    :to_resource   => 'compute',
    :attributes    => { "propagate_to" => 'both', "flex" => true, ""current" =>3, "min" => 1, "max" => 3}
end

schwankl:
@nzhang1220 lets do something like: https://github.com/oneops/circuit-oneops-1/commit/4f030464afbccb4b1c595d9b769a4f72d71aa896  so the lb is available in design

nzhang1220:
I did https://github.com/oneops/circuit-oneops-1/commit/4f030464afbccb4b1c595d9b769a4f72d71aa896 for `pack/lbdb.rb` and added the following into `pack/postgresql-governor.rb`, refreshed the pack files on the inductor, however it seems that `current`, `min`, `max` are still 1.

[ 'lb' ].each do |from|
  relation "#{from}::depends_on::compute",
  :only => [ 'redundant' ],
  :from_resource => from,
  :to_resource   => 'compute',
  :attributes    => { "propagate_to" => 'both', "flex" => true, "current" =>3, "min" => 1, "max" => 3}
end

nzhang1220:
@schwankl Comments? Should I simply upload the code so that you could visually check it?

schwankl:
yes, instead of lbdb level - override here: https://github.com/oneops/circuit-oneops-1/blob/master/packs/postgresql-governor.rb

PR #373, opened 2016-07-26
sensu in oneops

kaushiksriram100:
https://sensuapp.org/

Component will install and configure a sensu client for cloud ready monitoring.


components/cookbooks/sensuclient/metadata.rb
```diff
@@ -0,0 +1,103 @@
+name             'Sensuclient'
+maintainer       'sriramKaushik'
+maintainer_email 'kaushiksriram100@gmail.com'
+license          'All rights reserved'
```
schwankl: license should be Apache 2.0

PR #375, opened 2016-07-27
STROOSUP-559: adding support for nodejs pack with source and package …

ighouri:
…installation


schwankl:
this was already fixed in the node component you are trying to delete.  There is no reason to support old os-package versions of node...  Decision was to be binary only - verified already w/ Michael Fischer

schwankl:
see https://github.com/oneops/circuit-oneops-1/commit/a5598768b01de3f2fcf29ccbb160501f7edd2d48 ... this also allows newer versions to be used without code change

ighouri:
@schwankl 

1). I have removed the package and source installation of nodejs.

schwankl:
discussed w/ ighouri over slack - folks will continue to use node pack

PR #376, opened 2016-07-28
PG Governor: bump up default number of compute

nzhang1220:


nzhang1220:
@schwankl open a new PR, which refers to our discussion in https://github.com/oneops/circuit-oneops-1/pull/370. Not quite sure if this is the right direction, but would be happy to get your feedback.

packs/postgresql-governor.rb
```diff
@@ -265,11 +265,59 @@
...
+                ]
+            }
+            ]
+        }
+        ]
+    }'
 }
+}
+
+#resource "lb",
```
schwankl: rm old or add comment to why

nzhang1220:
@schwankl remove the old `lb` and hope this will look a little better.

nzhang1220:
@schwankl would it be nice if you may have a couple of minute to review the change? Thanks 

(I am just curious about learning how to correctly override a relation in the application pack level)

PR #377, opened 2016-07-28
Fixing DNS entries for individual instances

hemantb:


schwankl:
if its a hostname then will depend_on the compute ... no need to use RequiresComputes

PR #384, opened 2016-08-05
lb netscaler any all ; misc k8s

schwankl:


sleepyorion:
I've reviewed these changes internally before, these are the same changes.

PR #391, opened 2016-08-08
Vsphere compute

carlospuga:
Note: modification within the compute/recipes/base.rb
-- since I provided the capability for users to throttle the bandwidth for the rsync'ing of the recipes to the instances, it'll surpass the 10min default timeout for the Chef::Mixin::ShellOut.  Therefore I incremented the timeout to an hour (shell_timeout = 3600).
--currently the bandwidth throttle is with yum & rsync.  I will be changing this to be encompass all network traffic within network range.


schwankl:
https://github.com/oneops/circuit-oneops-1/pull/391/files#diff-7b99837c212cbc1140598dbd80737c86R106 bandwidth_throttle_rate should be renamed to rsync_throttle_rate

schwankl:
for os level iptables should be os_throttle_rate ... but imo this should be hash with interface key (ie limit eth0 by x and vpn0 by y)

schwankl:
lets make more generic:  ingress_rules (hashmap of names to iptable rules) and egress_rules (outbound rules)

schwankl:
will merge from now - lets rename attr and add ingress/egress_rules next iteration

PR #401, opened 2016-08-11
Making external raid as optional to the user than as default

umullan:


schwankl:
wrong way to impl - just add mode attribute - there is already support for it. https://github.com/oneops/circuit-oneops-1/pull/401/files#diff-048b8ae51c45bdb3478c48f5f65a6f59R301  if mode == no-raid then just pass the storage devices to lvm else it will pass the level [0,1,5,10] to mdadm.

schwankl:
closing due to was implemented using mode attr

PR #414, opened 2016-08-16
Adding a default pattern for daemon to prevent nill failures.

gauravlall:


schwankl:
instead of putting in a pattern that is not used, remove the pattern attr from being used on this line https://github.com/gauravlall/circuit-oneops-1/blob/2f25410674c6cd2e1861430f0bfb236b818b87a9/packs/tomcat.rb#L119

PR #431, opened 2016-08-25
Initial PR for panos firewall integration

kpgriffith:
Created a generic firewall cookbook then the specific implementation cookbook for Panos integration.

This only handles creating the addresses and dynamic address groups in the Panorama device.  It will not create or modify firewall rules at this time.


components/cookbooks/firewall/metadata.rb
```diff
@@ -0,0 +1,13 @@
+name             'Firewall'
+maintainer       'OneOps'
+maintainer_email 'support@oneops.com'
+license          'Apache License, Version 2.0'
+description      'Installs/Configures firewall'
+long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
```
schwankl: can you remove this long_description / dependency to README.md ... md and png files not rsync'ed and could fail when doesnt exist

PR #433, opened 2016-08-26
sleeping thread in between paritioning and formatting commands

carlospuga:


schwankl:
discussed - should not use these sleeps to attempt to work around race conditions... should be moved to volume::add ephemeral section like rest of providers

PR #436, opened 2016-08-29
adding cloud mirror service to tomcat component

ighouri:


schwankl:
fix branch conflicts.   php version default not good - who asked for that?

ighouri:
default version means to get this package from default repo in the VM (whatever version is available) in the VM. Otherwise other specific version, will come from the phpwebtatic repo

PR #439, opened 2016-08-30
code will no longer create sdb1. this will eliminate the need to slee…

carlospuga:
…p the thread while creating sdb1. it will also align with how the openstack and azure device are mounted.


carlospuga:
--no longer needed, this code will be replace by pull request #442

PR #454, opened 2016-09-08
increase wait time for windows computes to come online

kpgriffith:
60 seconds just isn't long enough for windows computes to come online.  It fails 1 to 2 times and eventually succeeds, so I increased it to 150 seconds.

### DIFF
--- a/components/cookbooks/compute/recipes/base.rb
+++ b/components/cookbooks/compute/recipes/base.rb
@@ -44,7 +44,7 @@ ruby_block 'install base' do
     end
 
     # add repo_list from os
-    if node.has_key?("repo_list") && !node.repo_list.nil?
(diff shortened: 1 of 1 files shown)


components/cookbooks/compute/recipes/base.rb
```diff
@@ -44,7 +44,7 @@
     end
 
     # add repo_list from os
-    if node.has_key?("repo_list") && !node.repo_list.nil?
+    if node.has_key?("repo_list") && node.repo_list.size > 0
```
schwankl: it can be nil and would fail. pull latest for tmp fix adding .include? '['

kpgriffith: @schwankl i pulled the latest.  I don't see this as part of the PR.

PR #483, opened 2016-09-27
Cookbook for Azure securtity groups

Guzabio:
Refactor for have an Azure Security Group cookbook


schwankl:
if secgroup will include azuresecgroup add that to secgroup metadata.   Soon we will only rsync the cookbook/components in the runlist and their metadata depends

Guzabio:
I added the depends to the secgroup metadata and I made tests.

PR #496, opened 2016-10-11
get rabbitmq password from

kaushiksriram100:
feature enhancement: request password for rabbitmq endpoint from customer

### DIFF
--- a/components/cookbooks/sensuclient/metadata.rb
+++ b/components/cookbooks/sensuclient/metadata.rb
@@ -53,6 +53,17 @@ attribute 'endpoint',
             :order    => 1
           }
 
+attribute 'sensu_vhost_password',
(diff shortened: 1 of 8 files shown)


schwankl:
is vhost a rabbitmq concept - seems to overlap with vmware vhost and if possible lets not mix

kaushiksriram100:
@schwankl vhost is rabbitmq concept in this case. vhost for rabbitmq.

components/cookbooks/sensuclient/metadata.rb
```diff
@@ -53,14 +53,25 @@
             :order    => 1
           }
 
+attribute 'sensu_vhost_password',
+          :description => 'vhost pwd',
```
kaushiksriram100: will do that.

PR #498, opened 2016-10-12
replace pwd to password

kaushiksriram100:


kaushiksriram100:
@schwankl can you please help with this merge?

components/cookbooks/sensuclient/metadata.rb
```diff
@@ -93,20 +64,184 @@
...
+
+attribute 'pagerduty_channel',
+          :description => 'pagerduty channels to forward alerts. Only one channel accepted.',
+          :format      => {
+            :help     => 'pagerduty channels to forward alerts. Separate multiple channles with comma. Will work only if you add pagerduty handler above.Make sure your key is already provided to esm-monitoring@walmart.com',
+            :category => '4.alarm notification options',
+            :order    => 5
+          }
+
+attribute 'memory_warning_threshold',
```
schwankl: all these metrics_severity_threshold attrs should be replaced by 1 thresholds hash attr with key something like memory_warning value 90, etc

PR #518, opened 2016-10-27
changes to support multiple volume type for cinder storage.

kowsalyapalaniappan:
changes to support multiple volume type for cinder storage.

### DIFF
--- a/components/cookbooks/storage/metadata.rb
+++ b/components/cookbooks/storage/metadata.rb
@@ -41,12 +41,14 @@ attribute 'volume_type',
   :required => "optional",
   :default => "GENERAL",
   :format => {
-    :help => 'Provide volume type',
+    :help => 'select the storage type. Storage type determines the volume type(IOPS) and service level(bandwidth)',
     :category => '1.Configuration',
     :order => 3,
     :form => { 'field' => 'select', 'options_for_select' => [
-      ["General", "GENERAL"],
-      ["IOPS", "IOPS"]] }
+      ["General", "GENERAL"], # default
+      ["Standard-1", "iops-low-300"],
+      ["IOPS-1", "iops-high-3500"],
+      ] }
   }   
 # maps provider vol-id for md
 attribute 'device_map',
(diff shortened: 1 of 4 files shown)


components/cookbooks/storage/recipes/node_lookup.rb
```diff
@@ -0,0 +1,37 @@
...
+  return true
+end
+
+# Get the volume type from the volume type map
+volume_type_selected = node.workorder.rfcCi.ciAttributes["volume_type"]
+Chef::Log.debug("node_lookup volume_type_selected: #{volume_type_selected}")
+if volumetype_map[volume_type_selected] != nil
+   node.set[:volume_type_from_map] = volumetype_map[volume_type_selected]
+else
+   puts "***FAULT:FATAL=Volume Type #{volume_type_selected} Not found ***"
```
kfilipovski: What happens for providers that don't implement a map?  This code should be in the cinder cookbook, not the storage cookbook.  The provider can choose to use fixed mappings and it will not break designs with storage component.

components/cookbooks/storage/metadata.rb
```diff
@@ -39,14 +39,15 @@
...
+  :default => "STANDARD1",
   :format => {
-    :help => 'Provide volume type',
+    :help => 'Select the storage type.Storage_types are mapped against volume types offered by cloud providers - see provider documentation for details. Storage type determines the volume type(IOPS) and service level(bandwidth).',
     :category => '1.Configuration',
     :order => 3,
     :form => { 'field' => 'select', 'options_for_select' => [
-      ["General", "GENERAL"],
-      ["IOPS", "IOPS"]] }
+      ["Standard-1", "STANDARD1"],
```
wwang001: Correct. These are our first guesses at service offerings, and it's possible that we may add more in the future, as we get feedback from users.

components/cookbooks/storage/recipes/node_lookup.rb
kowsalyapalaniappan: @kfilipovski , node_lookup is called only if the provider is cinder in storage/add.rb recipe. Its not executed if the provider is azure. I referred to node_lookup for compute sizes and its in compute cookbook, not in openstack cokbook. in future if other providers like azure provided multiple service levels they can customize the volume type map in the cloud templates.

Issue #521, opened 2016-10-27
Tomcat Pack Deployment fails when tomcat-daemon component added initially.


lkumili:
Issue Details : With tomcat pack if we configure tomcat-daemon as part of initial design, then deployment fails.

Below is the error from log: 
ERROR: service[tomcat7](daemon::add line 42) had an error: Chef::Exceptions::Exec: /bin/systemctl enable tomcat7 returned 1, expected 0

Root Cause : As part of deployment plan tomcat-daemon was deploying before tomcat installed on the vm.

Solution : Adding dependencies for tomcat-daemon  component on tomcat component and reloading the pack should resolve the issue.

PR #522, opened 2016-10-27
Update tomcat.rb

lkumili:
https://github.com/oneops/circuit-oneops-1/issues/521


lkumili:
Added tomcat as pre-requisite for tomcat-daemon component as part of tomcat pack change .

Deployed and validated the changes in local after reloading tomcat pack

bbourqu:
Very good catch.  We have this in tomcat-2[https://github.com/oneops/circuit-oneops-1/blob/master/packs/tomcat-2.rb#L266] but I guess we missed it in tomcat.

PR #534, opened 2016-11-02
using public IP address for the postgresql master

kfilipovski:
When using the private IP address the replication setup doesn't work when using multiple providers.


kfilipovski:
need to make this configurable via an attribute

PR #556, opened 2016-11-10
Update to FQDN to add DNS names inside the platform

dmontro:
Added a case to the fqdn component to allow for specifying the DNS component used.  If there is not an OS component in the dependencies, the first name in the aliases is used.

The intent of this update is to allow an additional DNS component to be added to the FQDN of an assembly for more complex packs that deploy multiple components.  For example, in Spark, a cluster may be deployed using the compute stack, but then it would be useful to have a separate compute for clients to connect to.  In this case, if the cluster FQDN is:

spark.dev1.myassembly.myorg.mycompany.com

...then this change will allow the creation of a client compute with the name:

**client**.spark.dev1.myassembly.myorg.mycompany.com

In order to use this update, a pack developer would create a fqdn component in their pack, then create a dependency to a compute or some other component that provides the **dns_record** attribute, but do **not** create a dependency on an OS component (which would ordinarily provide a host name).  This configuration results in an error in the current code.  With this patch, this configuration will cause the first alias to be selected for the DNS name that is generated.

### DIFF
--- a/components/cookbooks/fqdn/recipes/build_entries_list.rb
+++ b/components/cookbooks/fqdn/recipes/build_entries_list.rb
@@ -57,13 +57,31 @@ else
     fail_with_error "unsupported usecase - need to check why there are multiple os for same fqdn"
   end
   is_hostname_entry = true
-  ci = os.first
+  if os.size == 0
+    # When no OS is specified, the fqdn recipe cannot get the host name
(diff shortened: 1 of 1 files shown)


schwankl:
discussed with Daniel ; created PR for doing this using the manifest/RealizedAs ci name instead of aliases

PR #564, opened 2016-11-18
Added ntp service in iis and dotnetframework packs

sahebjade:
ntp service is required in windowsos cookbook

### DIFF
--- a/components/cookbooks/dotnetframework/recipes/update.rb
+++ b/components/cookbooks/dotnetframework/recipes/update.rb
@@ -1,0 +1,1 @@
+include_recipe 'dotnetframework::install'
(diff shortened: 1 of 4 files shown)


schwankl:
attributes should be the only thing overridden on the os component to set ostype.  This can all be removed:
https://github.com/sahebjade/circuit-oneops-1/blob/7e63ddd0cc813d31b0b1ad8ee6edf23a844bdc4d/packs/dotnet-framework.rb#L35-L40
... lets also remove:
https://github.com/sahebjade/circuit-oneops-1/blob/7e63ddd0cc813d31b0b1ad8ee6edf23a844bdc4d/packs/dotnet-framework.rb#L24-L25
https://github.com/sahebjade/circuit-oneops-1/blob/7e63ddd0cc813d31b0b1ad8ee6edf23a844bdc4d/packs/dotnet-framework.rb#L28-L31

PR #572, opened 2016-11-28
renamed iis-website to iiswebsite, in diagram(design) iis website image was not getting displayed…

sahebjade:
Looks like oneops don't like hypen (-) in the component name.


schwankl:
root cause should be found instead of workaround like this.

PR #590, opened 2016-12-12
removed tomcat 6 option

jksgit:


schwankl:
why is this being removed?

jksgit:
I was in wrong direction, Followed up with gourav.

Will submit a new pull request, Thanks Mike

PR #612, opened 2016-12-29
Replace REDIS bug fix

ravivangara17:


components/cookbooks/ring/recipes/replace.rb
```diff
@@ -1,2 +1,18 @@
...
+deps.each do |dep|
+  class_name = dep['ciClassName'].split('.').last
+  db_type = dep['ciClassName'].split('.').last.downcase
+  if class_name == "Redisio"
+    isRedis = true
+    include_recipe "ring::replace_#{db_type}"
+  end
+end
+
+if !isRedis
```
schwankl: ring::add still gets called on line 18

components/cookbooks/ring/recipes/replace.rb
```diff
@@ -1,2 +1,18 @@
+deps = node.workorder.payLoad.DependsOn
+
+db_type = nil
+isRedis = false
+deps.each do |dep|
+  class_name = dep['ciClassName'].split('.').last
+  db_type = dep['ciClassName'].split('.').last.downcase
+  if class_name == "Redisio"
```
schwankl: why not use db_type to match on? its needed to generate the recipe name

components/cookbooks/ring/recipes/replace.rb
ravivangara17: This is a check for REDIS replacement, if it is not redis (might be a cassandra replace) then it calls redis.add.

It works in MAIN-1 circuit

components/cookbooks/ring/recipes/replace.rb
```diff
@@ -1,2 +1,16 @@
+deps = node.workorder.payLoad.DependsOn
 
-include_recipe "ring::add"
+db_type = nil
+is_redis = false
+deps.each do |dep|
+  class_name = dep['ciClassName'].split('.').last
+  db_type = dep['ciClassName'].split('.').last.downcase
+  if class_name == "Redisio"
+    isRedis = true
```
schwankl: you didnt change isRedis here.  also, lets use db_type only and get rid of class_name

PR #629, opened 2017-01-13
Default Postgres archive_mode to off

kevinconaway:
@schwankl I'd like to have these changes merged to the postgres pack.  Please let me know your thoughts.

1. I have defaulted `archive_mode` to off.  I believe that this should not be on by default without some kind of plan in place to deal with the archived WAL segments.  Otherwise, users will be surprised by their disks filling up with archived segments that they didn't ask for

1. I left the `archive_command` in place so that it will work as designed should the user want to turn `archive_mode` on.

1. I moved both of those statements to the top level hash that gets merged with the user specified postgres conf keys.  This allows the user to override them if needed.

1. I removed the `archive_timeout = 0`  as this appears to be the default value.


kevinconaway:
@schwankl do you have time to review this? Or is there another person on your team who can?

PR #634, opened 2017-01-19
Added an extra wait for windows VMs at boot time

dschelchkov:


schwankl:
disgusting having to add these waits - but verified even on azure the initial boot time is 7min for their base image

PR #644, opened 2017-01-24
Additional Modifications to Tomcat 2

bbourqu:
* changed password generation to SecureRandom
* removal of old service before adding new service(so that version change is not so bad)
* defaulted artifact to false for `Expand`ing
* removed 8.5.6 replaced with 8.5.8
* changed shared_download_http back to remote_file(bug in parallel downloading)


bbourqu:
Thanks @schwankl for the merge!

PR #681, opened 2017-02-09
[READY FOR REVIEW] DO NOT MERGE until after tests are complete Tomcat 75 build change

bbourqu:
This change adds propagation to the artifact for tomcat as well as makes build version 75 available and the default


bbourqu:
I'll work the conflicts after testing I know they revolve around the HTTP verbs

PR #688, opened 2017-02-13
Removing the dependancy between objectstore and user components

umullan:
There is no need for objectstore depends on user. Hence removing it.

### DIFF
--- a/packs/base.rb
+++ b/packs/base.rb
@@ -762,7 +762,8 @@ end
   { :from => 'artifact',    :to => 'os' },    
   { :from => 'sensuclient', :to => 'compute'  },
   { :from => 'library',     :to => 'os' },
-  { :from => 'objectstore',  :to => 'compute'}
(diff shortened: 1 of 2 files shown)


packs/base.rb
```diff
@@ -762,8 +762,7 @@
   { :from => 'artifact',    :to => 'os' },    
   { :from => 'sensuclient', :to => 'compute'  },
   { :from => 'library',     :to => 'os' },
-  { :from => 'objectstore',  :to => 'compute'},
-  { :from => 'objectstore',  :to => 'user'}
```
schwankl: need this relation so that the user is created first

umullan: objectstore doesnot have any dependancy with user, all it does is to copy the openrc under /etc and enable the objectstore cli for endusers.

umullan:
Added the relation back in base pack. Please review and approve.

PR #700, opened 2017-02-18
Modify sensu rpm mirror (update add.rb)

kaushiksriram100:
modifying sensu rpm endpoints.


components/cookbooks/sensuclient/recipes/add.rb
```diff
@@ -100,8 +100,8 @@
 	when "rhel"
 		sensu_client_version = node[:sensuclient][:sensu_client_version]
 		system = `uname -m|tr --delete '\n'`
-		mirror = "https://sensu.global.ssl.fastly.net/yum"
-		rpm_package = sensu_client_version+"."+system+".rpm"
+		mirror = "https://sensu.global.ssl.fastly.net/yum/6"
```
schwankl: should be based on rhel/centos ... oneops-1 supports rhel/centos 7 ... so should be 7,el7

PR #717, opened 2017-03-01
READY FOR REVIEW Added policies on Tomcat version number

bbourqu:


bbourqu:
I see an issue with Tomcat 8 it should be `ciClassName:("bom.Tomcat" OR "bom.oneops.1.Tomcat-2")`

bbourqu:
Thank you for the merge @schwankl!

PR #727, opened 2017-03-04
adding cloud mirror service

ighouri:


sleepyorion:
@ighouri :   in the add recipe, mirrors itself is not define until L49.  Assuming the use case where we have created mirrors but not "maven" itself, then L50 would fail.  I think we should check for the existence of mirrors['maven'] before using it, if it's not there you can assign nil which you would check again at L56.  For L55 we assume that mirrors['maven'] is defined in L49, but in the case where it goes in condition L52 it would be an empty hash L55 again would fail because the key 'maven' doesn't existed.

sleepyorion:
@ighouri : something like this _source_list = mirrors.key?('maven') ? mirrors['maven'] : nil would work

ighouri:
@sleepyorion extra checks not needed. it will be nil if maven is not defined in cloud mirror service.

PR #735, opened 2017-03-09
improved inline docs for compute

mosabua:


mosabua:
Can I just merge this since its trivial @lkhusid @schwankl @vitaliyzinchenko  ?

PR #760, opened 2017-03-18
adding standard output message

ighouri:


schwankl:
print stderr too - that would allow easier debugging

PR #766, opened 2017-03-22
Created a pack for a Big Data client

dmontro:
This pack can be used as a client for the YARN and Spark packs.


dmontro:
Note:  This pack does have a dependency on the PR for Centrify, which is pull request #767 .

PR #817, opened 2017-04-26
Bug: Ansible cookbook metadata name update

hrodz13:
the command 'knife pack sync' is failing because the pack and the cookbook do not have the same name case-wise.

https://github.com/oneops/circuit-oneops-1/blob/master/packs/ansible.rb#L3


hrodz13:
I'm not sure which one (cookbooks/ansible/metadata.rb or packs/ansible.rb) should change, since I see that it has been back and forth. As long as they have the same case.

dmontro:
Capital letters in the component names is the pattern I have been following.  I remember hitting the stumbling block related to case in component names.

PR #819, opened 2017-04-27
Revert "Bug: Ansible cookbook metadata name update"

dmontro:
Reverts oneops/circuit-oneops-1#817


hrodz13:
Why was this reverted?

PR #848, opened 2017-05-18
Allow User to Specify Max Connection on Vip.

kowsalyapalaniappan:
Allow User to Specify Max Connection on Vip in software load balancers.


ighouri:
@kowsalyapalaniappan is this LB nuetron service for an additional cloud LB service other than netscaler ?

kowsalyapalaniappan:
yes, its software load balancer.

PR #854, opened 2017-05-29
pack changes for mssql monitors & start/stop/restart of mssql

rrs0108:


rrs0108:
will be raising another PR with new changes included

PR #874, opened 2017-06-09
Latest changes to mssql pack & component

rrs0108:
- Some corrections to mssql pack
- Component changes to add sysadmins & modify sa pwd post deployment
- Auto start the agent service after deployment
- Some changes to start/stop/restart on demand recipes


components/cookbooks/mssql/recipes/update.rb
```diff
@@ -2,4 +2,30 @@
 # Cookbook Name:: mssql
 # Recipe:: update
```
dschelchkov: Please move this functionality to add recipe. There are scenarios when after updating a component it will be an add or replace recipe that'll run, not the update. So this code has to be idempotent and reside in add recipe. It can be in a separate rb file, but essentially it should be called from add.rb

rrs0108:
Sure. will raise a new PR. closing this one

PR #881, opened 2017-06-13
Removing AppVersion and debug procedures from Tomcat v1 pack

bbourqu:


bbourqu:
Thank you so much @sleepyorion for the merge!

PR #889, opened 2017-06-22
check if nic is already available on update and get the same nic

shantanukande:


shantanukande:
to take care of subnet full error, checking subnet for available ip and using same. Take care of 'NIC deletion from azure side', catching exception and displaying message.

PR #891, opened 2017-06-23
User

ighouri:


sleepyorion:
@ighouri : is there a reason why we are changing home directory in the same PR as dependency fixing?  In the future please refrain from combining too many changes in same PR.

ighouri:
@sleepyorion yup aware. seems  i had some issue while creating new branch. i did git push in my forked repo and now it seems fine an dhave changes related to home app directory only.

sleepyorion:
@ighouri : Please explain the reason for changing home directory to another path instead of leaving the way it is originally.

ighouri:
@sleepyorion this is to prevent user home directory (/app) with volume mount directory (/app). User complains that their public keys gets lost after the volume component (as this recreate this /app again) and then they have to touch update user component to redeploy their keys again.

PR #893, opened 2017-06-23
update the netscaler saveconfig warning

mtunganati:


mtunganati:
@sleepyorion reson for removing this is because netscaler can only do one save at a time and because we have multiple deployments happening all the time so save config is failing after sometime which is causing deployment failures. later some other deployment will save the config and all will be good.The only downside of this change is when netscaler itself crashes before some other deployment actually does a proper save then the changes will be lost. please review and see if this is reasonable ask to not.

mtunganati:
will create a separate PR

PR #895, opened 2017-06-27
Revert "Add an attribute for owner email"

ighouri:
Reverts oneops/circuit-oneops-1#868


ighouri:
Spoke with Jayesh. This does not needs to be merged as he will create a separate PR with other changes to be merged.

PR #907, opened 2017-07-05
Fix ruby block by adding only_if condition.

jhescobar05:
Fix ruby block by adding only_if condition.


dschelchkov:
Can be merged if CLA is signed

PR #913, opened 2017-07-10
Add the ability to specify a file mode for created files

kevinconaway:
@jayeshpise This PR adds the ability to specify a file mode for files that are pushed to the system.

Currently all files are created with _0755_ which isn't always what we want.  I realize its possible to do this with the _exec_cmd_ but it gets ugly if you want to run other commands as well.


jayeshpise:
@kevinconaway, looks good, merging. You can also consider adding a regex to make sure the value entered for this new attribute is a valid one.

PR #920, opened 2017-07-14
Adding support for multiple storage components

dschelchkov:


dschelchkov:
That's by design. Basically I need a user's input (or the pack's owner can set it at the pack level) for that attribute. If the value for that attribute is not provided (which is what's gonna happen now for all the existing packs), then the 1st branch of the matching algorithm will not be used. The 3 "branches" are as follows:
1) if "Based On Storage" is not empty - use its value to find matching storage component
2) if #1 didn't yield a match - compare the name of the volume components to all storage component names, in this manner: 'volume-foo' matches 'storage-foo', but "volume-foo" won't match "storage-foobar" or "barstorage-foo" etc
3) if neither #1 or #2 worked, use the old approach of using the first storage from the array of available storage components

PR #929, opened 2017-07-21
Barbican : Correcting FOG import

ganpatagarwal:


ighouri:
@ganpatagarwal confirmed that fog required module works in a dev environment. merging it

PR #931, opened 2017-07-25
docker pack : default volume will not change to storage when storage is added

shantanukande:
default volume component will be dependent on compute/OS and will not change to storage when storage is added. If storage is added and then another volume is added, those will depend on storage.


sleepyorion:
@shantanukande : why leave it commented?  We can always go back in git history to find previous change if desired.  It just doesn't look right with those being commented out without any explanation whatsoever.  So it's best to just remove it.

PR #940, opened 2017-07-28
Add Kitchen CI tests for NODE

ravivangara17:


ravivangara17:
Remove the "cable connected line" from kitchen Yaml file

PR #944, opened 2017-08-07
Don't merge: user can able to pull any revision from git

shantanukande:
Override the clone method which will now consider revision mentioned by user while cloning repository.


shantanukande:
Changes are done for delete recipe and added kitchen-ci test cases for add, update and delete operation.

PR #953, opened 2017-08-15
lb-cert autogen related change to use ciBaseAttributes in case of add-rfc

jayeshpise:
While uploading lb cert to netscaler, use ciBaseAttributes of the payLoad.DependsOn.certificate if the lb-cert is auto-generated and is an add rfc


sureshg:
LGTM

PR #954, opened 2017-08-16
Resolving json conflit for ubuntu-16.04

shantanukande:
For Ubuntu-16.04, remove two out of three json gem so that it will not create conflict for azure gems


shantanukande:
Reverted changes in volume and deleting json gems which are creating conflict at OS level.

PR #956, opened 2017-08-16
tomcat pack updated to work latest version

msumankumar:


bbourqu:
Discussed with @msumankumar and I understand the change as well as the scenario. @jwood8 please make sure you pull this in with your changes to ensure we are testing against this change.

PR #974, opened 2017-09-05
Don't merge: Azuredisk attach onupdate

shantanukande:
Check if storage_provider is azure and if it is, it will attach storage to compute. This is helpful for main and labs circuit as storage can't be attached on compute.


dschelchkov:
volume and storage component have been re-factored

PR #975, opened 2017-09-06
Adding additional check for FQDN recipe with deploy on failure flag enable

mtunganati:


components/cookbooks/fqdn/recipes/build_entries_list.rb
```diff
@@ -123,6 +123,11 @@
 deps = node.workorder.payLoad[:DependsOn].select { |d| d[:ciAttributes].has_key? "dns_record" }
 values = get_dns_values(deps)
 
+# check if dependent component creation is a success or else fail the reciepe execution
+if values.nil? || values.empty?
+   exit_with_error "Empty dns_record. Please check whether the depends on component of FQDN sucessfully completed"
```
sureshg: @mtunganati instead of `exit 1`, should we add a proper error for this case also  - https://github.com/oneops/circuit-oneops-1/blob/master/components/cookbooks/fqdn/libraries/fqdn_base.rb#L220-L222 ? Do we need  `depends on component ` in the error message displayed to end user?

mtunganati: yes @sureshg we need to have that message because that will help the team to check if the depends on executed and successful

sureshg: LGTM. Merging.

components/cookbooks/fqdn/recipes/build_entries_list.rb
```diff
@@ -123,6 +123,11 @@
 deps = node.workorder.payLoad[:DependsOn].select { |d| d[:ciAttributes].has_key? "dns_record" }
 values = get_dns_values(deps)
 
+# check if dependent component creation is a success or else fail the reciepe execution
+if values.nil? || values.empty?
+   exit_with_error "Empty dns_record. Please check whether the compute/lb deployment step passed successfully"
```
sureshg: LGTM. Merging.

PR #976, opened 2017-09-07
Added propagate_to for tomcat-daemon to certificate

bbourqu:
Allows the daemon to restart tomcat when the certificate is changed(usually due to a certificate renewal)


jayeshpise:
Looks good to me. This propagation makes sense.
@bbourqu @gauravlall , as a side note, we should make tomcat-daemon as mandatory. Currently it is optional in this pack

bbourqu:
Thanks @jayeshpise for the merge!

PR #979, opened 2017-09-08
script changes for installation of rvm, ruby 2.3.3, rubygems 2.16.12 on the compute

rameshjanjyam:
for azure, the volume add recipe - https://github.com/oneops/circuit-oneops-1/blob/Azure-Refactor-testing/fog-arm-migration/components/cookbooks/volume/recipes/add.rb#L47 - executes `azuredatadisk::attach` recipe for attaching the disk. This `attach` recipe uses `fog` and makes azure calls to attach the disk. 

This volume recipe is executed on the compute itself. since it depends on `fog-azure-rm` the compute needs to have `ruby 2.3.3` on it. 
on these `centos` boxes, as of now, the only way to get `ruby 2.3.3` seems to be via `rvm`. so installing `rvm` first and using it `ruby 2.3.3` and `rubygems 2.6.12` are installed. 

If the `yum repos` on these centos computes have `ruby 2.3.3` then installing ruby via `yum install` would be the better solution for installing `ruby 2.3.3` on compute


gauravlall:
@rameshjanjyam  Please remove , we need to externalize the walmart specific urls

PR #981, opened 2017-09-11
Netscaler mssql - DO NOT MERGE

dschelchkov:


dschelchkov:
We don't want to merge this now, but this new functionality has to be preserved. At some point we might get back to it. So closing the PR but keeping the branch.

PR #994, opened 2017-09-20
Tmpfs support for storing secrets on computes.

sureshg:
Changes in this PR.

1.  Make `/secrets` the default non-changeable secrets mount point. This makes the tmpfs mount point management a lot easier and provides consistent access to end-user applications. `Sqaure` also follows the same pattern.

2.  Make `tmpfs` mount point for secrets
      * For `systemd` systems, we make use of `Mount` unit files and make `keysync` service depends on it.
      * For SysV init systems, use normal `mount` command and made sure it's the part of service start command (check `mount_tmpfs` function). 

3. Keysync config changes to make sure the sync happens only for tmpfs (`TMPFS_MAGIC           0x01021994`) secrets mount (`filesystem_type `).

4. Fixed some ruby code style based on Rubocop recommendations.


gauravlall:
Looks Good to me.

PR #1019, opened 2017-10-03
update and fault domain based on region

dandagopi:
1. pass update and fault domain based on azure region
2. if region not defined we pass default 2


gauravlall:
Is there any better way to avoid hardcoding ? we can look at configuration attribute on cloud service

PR #1029, opened 2017-10-09
Integration test fix

bilal-naeem-confiz:
Made changes to fix integration tests


bilal-naeem-confiz:
These changes fix the following items:
- Made _idle_timeout_in_minutes_ an argument to the _build_public_ip_ method. The Public IP object attached to an Application Gateway should not have the _idle_timeout_in_minutes_ attribute set.
- Changed _add_ recipe for _azure_lb_ to retrieve the _network_interface_card_id_ from the VM.
- Added a check to see if a _record_set_ exists before deleting it.

PR #1035, opened 2017-10-12
calling gdns service for azure on basis of cloud priority

shantanukande:
GDNS service will now get called on cloud priority basis and if cloud priority is changed to secondary, we are unbinding gdns for that cloud. No code change for open-stack.


dandagopi:
@shantanukande  can you add Kitchen CI test cases please

PR #1045, opened 2017-10-19
Fixed azure cookbook unit tests

bilal-naeem-confiz:
Fixed azure cookbook unit tests


bilal-naeem-confiz:
Changes in this PR:
- Used fog-azure-rm `check_network_interface_exists` in `delete` method for NIC
- Created a `load_availability_set_response` method to get the availability set response. Previous implementation was causing the unit tests to break.
- Used `File.expand_path` so that the file can be accessed from any point in the codebase

PR #1053, opened 2017-10-19
Ensure that default value results in valid protocol

bbourqu:
screen shot 2017-10-18 at 09 55 44
screen shot 2017-10-18 at 09 54 40
screen shot 2017-10-18 at 09 54 12


bbourqu:
Thanks @gauravlall for the merge!

PR #1056, opened 2017-10-23
Updating certificate recipe to copy compute certs to compute

msumankumar:


components/cookbooks/certificate/recipes/add.rb
```diff
@@ -12,10 +12,27 @@
...
-	expires_on = node[:expiry_time]
+     run_context.include_recipe provider + "::add_certificate"
+     expires_on = node[:expiry_time]
+     if node.workorder.payLoad.has_key?("DependsOn")
+           cer_cont = node[:cert_content]
+           key_cont = node[:key_content]
+           ca_cont = node[:ca_content]
+           pfx_cont = node[:pfx_content]
+           `mkdir -p #{path}`
+           File.open("#{path}/cert_content.pem", "w") {|line| line.puts "\r" + cer_cont.to_s }
```
sureshg: IMO, no need for `content` in the file name. Usually, it's using the format
* `common-name.crt` or `.pem`
* `common-name.key`
* `ca_cert.crt` 
* `common-name.pfx`

components/cookbooks/certificate/recipes/add.rb
```diff
@@ -12,10 +12,27 @@
 end
 
 expires_on = node[:expiry_time]
-
+cer_cont = ""
+key_cont = ""
+ca_cont = ""
+pfx_cont = ""
```
sureshg: No need to init these vars. Remove them.

components/cookbooks/certificate/recipes/add.rb
```diff
@@ -12,10 +12,24 @@
 end
 
 expires_on = node[:expiry_time]
-
+cert_path = node[:certificate][:path]
+ci_name = node[:app_name]
+path = "#{cert_path}/#{ci_name}"
+cname = node.workorder.rfcCi.ciAttributes.common_name
 if !auto_provision.nil? && auto_provision == "true" && !provider.nil? && !provider.empty?
```
sureshg: Shouldn't we add the check for managed_via to make sure it's created on remote compute?

components/cookbooks/certificate/recipes/add.rb
```diff
@@ -12,10 +12,24 @@
...
-	include_recipe provider + "::add_certificate" 
-	expires_on = node[:expiry_time]
+     run_context.include_recipe provider + "::add_certificate"
+     expires_on = node[:expiry_time]
+       if node.workorder.payLoad.has_key?("DependsOn")
```
msumankumar: @sureshg Used this condition `if defined?(node.workorder.payLoad.ManagedVia)` for some reason this was not working so used "DependsOn"

jayeshpise:
@msumankumar : Lets make the persist cert logic conditional based on value node[:persist_cert]
Also, can you modify the secrets-client recipe accordingly to set that value to false so that the certs don't get saved on the compute in that case

gauravlall:
@msumankumar  We need kitchen ci test

msumankumar:
Added the kitchen CI test cases

PR #1098, opened 2017-11-08
added integration tests, for azure, to compute and lb cookbook.

rameshjanjyam:


sleepyorion:
@rameshjanjyam : can you remove the usage of kitchen_proxy.rb, and start using monkey_patch.rb.

rameshjanjyam:
@sleepyorion is there a cookbook that is already written using `monkey_patch.rb` i will verify it to know the changes required to use it. what has to be changed on local inductor to make use of this new setup? should the kitchen need to be reinstalled on local vagrant setup

rameshjanjyam:
@sleepyorion is there a cookbook that is already written using `monkey_patch.rb` i will verify it to know the changes required to use it. what has to be changed on local inductor to make use of this new setup? should the kitchen need to be reinstalled on local vagrant setup

PR #1100, opened 2017-11-09
[DO NOT MERGE]Removed deprecated values and set defaults for the tomcat pack

jwats56:


bbourqu:
@jwats56 you signed in as two different users.  Both would be need to sign the CLA.

PR #1111, opened 2017-11-14
Fixed fast deploy logic to use runtimes yaml

CodyLeeWhite:
Same thing as the first fast deploy update but cleaner and more dynamic.


bbourqu:
Can you ensure that if the zip file is not found that we fail gracefully and output a message?

CodyLeeWhite:
done

PR #1124, opened 2017-11-21
Fixed the gem conflict for net-ssh. Replaced the hard-coded gem source

dschelchkov:
Same approach we use for gemfiles in test suites.
Checking in /etc/oneops file first for rubygems key
If not found use whatever gem source already configured


dschelchkov:
Gemfile was moved to oneops repo

PR #1127, opened 2017-11-22
Local SSD Integration test

muhammad-asad-confiz:
Local SSD integration test written.


dschelchkov:
I see several issues with this PR:
1) the way the spec files are written these tests will run for every deployment that has volume component, including Openstack deployments, BM deployments, Platforms with persistent storage and not ephemeral, windows platforms etc. And these spec files assume the deployment to be for an ephemeral disk in azure and for linux only.
2) Why the use of SSH? this is a for a volume component, which is managed via compute, i.e. it runs on the VM already. The current code will result in VM opening ssh to itself
3) Describe 'local SSD' which asserts mount point and file system is a duplication of an existing volume assertion - see test/integration/add/serverspec/add_spec.rb line 11-21 
4) incorrect folder structure - test/integration/ssd will never be used by automated kitchenci tests. The correct format is test/integration/<recipe_name>  where <recipe_name> can be add, delete, update or replace.  Plus why using 2 files - ssd_spec.rb and then ssd.rb, what is the reason for such separation? 
I suggest moving everything into ssd_spec.rb file and place that file in test/integration/add/serverspec/ folder. Or even just adding the assertions to add_spec.rb file
5) this is just me nitpicking, can safely ignore this point, but I honestly don't understand the use of functions like this one: 
def get_ephemeral_file_system
   @node['volume']['fstype']
 end
What's wrong with just referencing node object directly, instead of this:
FS_TYPE = @spec_utils.get_ephemeral_file_system
use FS_TYPE = $node['volume]['fstype']
Saves few lines of code and in my opinion is easier to read

muhammad-asad-confiz:
We will be releasing a new PR soon

PR #1131, opened 2017-11-23
Updated Azure DNS Integration Tests via Kitchen CI

maham-nazir-confiz:
Azure DNS integration tests via Kitchen CI.

**Changes Made**
- **'zone.rb'** library from **'azuredns'** cookbook has been updated. The method call **'check_for_zone'** had a typo mistake. It was calling a Fog method that did not exist. That has been fixed.
- **'dns.rb'** library from **'azuredns'** has been updated for method call **'set_a_type_records'**. This method is responsible for creating and setting record set records per each deployment. The method did not set the records properly in case of scale up and scale down. That has been fixed.


maham-nazir-confiz:
@dandagopi 
Merge conflicts have been resolved

PR #1141, opened 2017-11-29
Object store support added for Azure

muhammad-asad-confiz:
- Support for object store has been added for Azure cloud.
- Moved from .erb template to yml file for optimization
- Added serverspec test cases
- Fixed large file-size download issue i.e. file-size > swap-size
- Objectstore listing, verbose and non-verbose flags support added


components/cookbooks/objectstore/metadata.rb
```diff
@@ -6,37 +6,7 @@
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
 version          '0.1.0'
 depends 	'swift'
+depends 	'azureobjectstore'
```
dschelchkov: we don't need dependencies on swift or azureobjectstore anymore, since we're not referencing recipes or libraries from those cookbooks.

muhammad-asad-confiz:
PR updated

PR #1178, opened 2017-12-11
Added a new component 'service-mesh'

vsrp1010:
1. Added a new component 'service-mesh' to deploy service mesh process for service registry. This component will be used by all user services of service registry. 
2. Added service-mesh component to base pack. This component is optional.
3. Added default ports (4140, 9990) to tomcat pack. These ports are needed for service-mesh process communication.


components/cookbooks/service-mesh/attributes/default.rb
```diff
@@ -0,0 +1,2 @@
+#Install
+default['service-mesh']['nexus-repo-url'] = "http://gec-maven-nexus.walmart.com/nexus/content/repositories/pangaea_releases/com/walmart/soa-linkerd"
```
jayeshpise: @vsrp1010 : I think the repo url should be externalized somewhow

jayeshpise:
@vsrp1010 : Added some inline comments on some of the files. Please check.

Also, i see you are starting your process by a command and that you do not have any init or systemd script/configs. This would mean if the compute is rebooted, your process will not start automatically.

packs/tomcat.rb
```diff
@@ -256,7 +256,7 @@
          :cookbook => "oneops.1.secgroup",
          :design => true,
          :attributes => {
-             "inbound" => '[ "22 22 tcp 0.0.0.0/0", "8080 8080 tcp 0.0.0.0/0", "8443 8443 tcp 0.0.0.0/0" ]'
+             "inbound" => '[ "22 22 tcp 0.0.0.0/0", "8080 8080 tcp 0.0.0.0/0", "8443 8443 tcp 0.0.0.0/0", "4140 4140 tcp 0.0.0.0/0", "9990 9990 tcp 0.0.0.0/0" ]'
```
bbourqu: Remove these from the tomcat pack.  Those that add the service mesh can add them, the pack is for tomcat not for service-mesh

vsrp1010: @bbourqu service-mesh component is proposed to be a mandatory component in near future and then these ports will be needed in all packs. As we are rolling out this component in a phased manner, I added them specifically in the pack which we are rolling out first. Could we keep them similar to other ports so that users don't have to add them every time they add service-mesh component?

bbourqu: Service-mesh is not mandatory for open-source end users.

vsrp1010: I'll revert this change.

bbourqu:
Thanks @vsrp1010 I approve of the removal of ports from tomcat.  I think @jayeshpise still has some review comments

vsrp1010:
@jayeshpise I have updated the PR including changes as per comments. Can you review again please?

vsrp1010:
@jayeshpise Could you take a look and merge this PR if everything looks good. All wm references are removed.

PR #1198, opened 2017-12-18
STRCASS-3619 Port apache_cassandra pack from circuit-walmartlabs-1

ajweave:
Corrected author so CLA will match (hopefully).


ajweave:
dupe

PR #1216, opened 2017-12-22
Added test to verify OS component

CodyLeeWhite:
It pulls file names, modes, owners , gem names, yum names from the checklist.yml file. With that there is 172 tests. More can be tested in network and proxies are not fully covered because It would take me to long right now to cover them. 

I had to refactor a section of OS(add-conf-files.rb) to make the test cleaner. I wrote the ugly version of the the test and then refactored it while testing against the ugly version. Then I cleaned the test up once it started passing.


components/cookbooks/os/test/integration/add/serverspec/checklist.yml
```diff
@@ -0,0 +1,213 @@
...
+  - chef
+  - mime-types
+  - ms_rest
+  - ms_rest_azure
+  - azure_mgmt_compute
+  - azure_mgmt_network
+  - azure_mgmt_storage
+  - azure_mgmt_resources
+  - azure
+  - rack
```
dschelchkov: See some duplicates here: chef, rack for instance..

components/cookbooks/os/test/integration/add/serverspec/centos_redhat_7_spec.rb
```diff
@@ -0,0 +1,131 @@
+require "/home/oneops/circuit-oneops-1/components/spec_helper.rb"
```
dschelchkov: If I understand correctly, this spec file will be executed for all platforms, even for windows, and in windows the require statement most likely will fail

components/cookbooks/os/test/integration/add/serverspec/centos_redhat_7_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    else
+      describe file(check['name']) do
+        it { should be_file }
+        it { should be_mode check['mode'] }
+        it { should be_owned_by check['owner'] }
+        it { should be_grouped_into check['group'] }
+      end
+    end
+  end
+
```
dschelchkov: So everything below this line will be run on all platforms?
That might cause problems, especially on windows 
I'd suggest moving this entire spec file outside of add/serverspec folder, and put add_spec.rb file back to add/serverspec folder
then in the add_spec.rb just check the platform, and if centos - require this centos_spec.rb

CodyLeeWhite:
fixed windows compatibility and duplicate gem entries.

PR #1224, opened 2017-12-27
Volume Kitchen CI tests fix

muhammad-asad-confiz:
Fixed volume_spec only returning free disk space. Now it returns the size of the attached disk.


dschelchkov:
@muhammad-asad-confiz 
Not sure what issue this PR is trying to solve, but it will not work correctly for the case where we have multiple volume components.
Please check out this PR https://github.com/oneops/circuit-oneops-1/pull/1228 , maybe it's already addressing whatever issue you're trying to solve here.

PR #1228, opened 2017-12-27
Fixed a bug with multiple volume groups

dschelchkov:
When persistent and ephemeral storage are used in the same pack
it will result in 2 volume groups created.
The code that determine VG size will pick arbitrary volume group.
Fixed by explicitly specifying VG when checking its size with vgdisplay

### DIFF
--- a/components/cookbooks/volume/test/integration/volume_spec.rb
+++ b/components/cookbooks/volume/test/integration/volume_spec.rb
@@ -50,7 +50,8 @@ else
   size_vm = `df -BG | grep #{$mount_point}| awk '{print $2}'`.chop.to_i
 end
 if !is_windows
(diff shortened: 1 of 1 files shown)


muhammad-asad-confiz:
@dschelchkov 

The volume component Kitchen CI tests are failing due to [this](https://github.com/dschelchkov/circuit-oneops-1/blob/a0c033b1423bc60acdad72891830933c12e1eeeb/components/cookbooks/volume/test/integration/volume_spec.rb#L48)
The problem will be solved by replacing the above mentioned line with the following line. 
```size_vm = `lvs --noheadings --units g #{lvm_dev_id} | awk '{print $4}'`.chop.to_i```

PR #1238, opened 2017-12-29
Busser gem issue resolved

muhammad-asad-confiz:
Resolved bug in install_base.sh when gem source was https://rubygems.org/.


dandagopi:
@rameshjanjyam  can you please review PR

PR #1239, opened 2017-12-29
Attach Managed Data Disk to VM

maham-nazir-confiz:
For Azure: Managed data disk for a VM was being created but not attached to the VM. Code has been added to attach the managed data disk to the VM.


dschelchkov:
@maham-nazir-confiz , in the future please discuss volume and storage issues with me before submitting a PR
Attaching of disks happens in volume component, not in storage

PR #1240, opened 2017-12-29
Changed Systemd to Initd

adeetikaushal:


sleepyorion:
@adeetikaushal : qq, does the installer drop down startup in init.d or is it systemd style?

adeetikaushal:
It is init.d style.

PR #1241, opened 2017-12-29
Fixing OS test's fix

CodyLeeWhite:
https://github.com/oneops/circuit-oneops-1/pull/1233

that pull request updated the modes without taking into account why they needed to change the modes. So I put in a hard chmod to make sure that the modes are constant for anyone that runs OS.

### DIFF
--- a/components/cookbooks/os/recipes/add.rb
+++ b/components/cookbooks/os/recipes/add.rb
@@ -158,7 +158,10 @@ else
 end
 
 # sshd
-execute "cp /etc/ssh/sshd_config /etc/ssh/sshd_config.backup"
+file '/etc/ssh/sshd_config.backup' do
+  mode 0644
+  content ::File.open('/etc/ssh/sshd_config').read
+end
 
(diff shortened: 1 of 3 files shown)


sleepyorion:
@CodyLeeWhite : in those commands, do you really meant to use | as in pipe which sends output from one command to another?  Or do you really mean to use && which ensure the previous command success before execute subsequent command?

muhammad-asad-confiz:
@CodyLeeWhite 

**Issue**
- OS Component Kitchen CI tests are failing due to file mode for `/etc/cloud/cloud.cfg`.

**Observation**
- The Kitchen CI test checking for the file mode for `/etc/cloud/cloud.cfg` to be 0664 [here](https://github.com/oneops/circuit-oneops-1/pull/1241/files#diff-e32209f792b97df0100b73da79735b28R174).

**Solution**
- We should set the file mode of the `/etc/cloud/cloud.cfg` file to 0664 at the time of creation.

PR #1256, opened 2018-01-08
Fixed rails to make it compatible with kitchen 

CodyLeeWhite:
Also fixed yum group install to be skipped if using a walmart repo.


sleepyorion:
is there any other way besides explicitly looking for walmartism ?

PR #1264, opened 2018-01-09
removed managedVia compute for hostname

sbonde:


sbonde:
managedVia compute on hostname is not required as hostname recipe don't need to be run on compute. In base pack it is not managedVia compute and it was mistakenly added as managedVia compute in solrcloud pack

PR #1265, opened 2018-01-10
nc command changed to remove -z 

sbonde:
'nc -z host port' command doesn't work on centos 7 as there is no -z option.
Changed the command to 'nc host port </dev/null'


components/cookbooks/solrcloud/templates/default/check_solr_zk_conn.sh.erb
```diff
@@ -35,7 +35,7 @@ log_echo "Total number of available ZK nodes = ${#zk_list[@]}"
 
 ping_zk () {
   zk=$1
-  (nc -z $1 2181)
+  (nc $1 2181 </dev/null)
```
sgoyal6: why are we reading from /dev/null ?

sbonde:
< /dev/null is used to instantly send EOF to the program, so that it doesn't wait for input
this is replacement of -z

PR #1268, opened 2018-01-10
Took all ambiguity out of file modes in os and compute

CodyLeeWhite:


CodyLeeWhite:
Don't merge yet. Finishing up final test.

PR #1269, opened 2018-01-10
Took all ambiguity out of file modes in os and compute

CodyLeeWhite:


CodyLeeWhite:
I have added the tests that were requested

CodyLeeWhite:
Also fixed the bind package install error that only happened on azure.

PR #1298, opened 2018-01-19
Added azuredatadisk dependency in azure cookbook metadata

bilal-naeem-confiz:
Added `azuredatadisk` dependency in `azure` cookbook metadata to fix `azure::del_blob` issue (Unable to find DataDisk)


dschelchkov:
azuredatadisk has been retired and should not be used anymore

PR #1307, opened 2018-01-24
Artifact sync

CodyLeeWhite:
Copied over the chef rest file from shared into artifact. 
Also wrote a unit test for the calculate parts method.


CodyLeeWhite:
If you want to try out the unit test just cd to the /unit/ dir and run rspec

PR #1326, opened 2018-01-30
Strcrdbmsd 1220

BrandonBenge:
Fixing the following issues
https://jira.walmart.com/browse/STRCRDBMSD-1644
https://jira.walmart.com/browse/PLCRDBMS-2018
and part of https://jira.walmart.com/browse/STRCRDBMSD-1502


components/cookbooks/cloudrdbms/recipes/restart.rb
```diff
@@ -1,4 +1,132 @@
...
+log "CloudRDBMS concord: start ansible job"
+### Start Concord Job for ansible ####
+bash 'start_ansible' do
+    cwd '/app'
+    user 'app'
+    code <<-EOF
+      # Set my needed qualities to call ansible
+      mv "#{playbook}.yml" _main.yml
+      curl  -u "#{managedserviceuser}:#{managedservicepass}" -F request=@_main.yml -F org="Default" -F project="cloudrdbms" -F repo="#{cloudrdbmspackversion}" -F entryPoint="ansibleFlow" http://#{concordaddress}/api/v1/process >./curl_#{playbook}.out
+      RC=$?
```
sjangra-git: Should we have a retry logic based on HTTP response code (NOT 200) or something for this curl command? In this recipe and others like the add

components/cookbooks/cloudrdbms/recipes/restart.rb
```diff
@@ -1,4 +1,132 @@
...
+
+log "CloudRDBMS concord: Wait for job to exit concord job"
+# Wait for job to exit concord job
+ruby_block 'validate_complete' do
+    block do
+      curlAttempt=1
+      curlComplete=0
+      while (curlComplete == 0) && (curlAttempt < 360) do
+          begin
+            status=`export playbook="#{playbook}" concordaddress="#{concordaddress}" managedserviceuser="#{managedserviceuser}" managedservicepass="#{managedservicepass}" && /app/get_concord_status.sh`
```
sjangra-git: Minor: Wherever possible we should try to keep one programming language as possible, so the readability of code is good. Like in this ruby block we have call to a shell script. This is general comment for future work so don't try to fix everything. We had some similar snippets in the legacy code.

components/cookbooks/cloudrdbms/recipes/stop.rb
```diff
@@ -1,30 +1,100 @@
...
+log "CloudRDBMS Show environment for managedserviceuser: '#{managedserviceuser}'"
+log "CloudRDBMS Show environment for clustername: '#{clustername}'"
+log "CloudRDBMS Show environment for drclouds: '#{drclouds}'"
+log "CloudRDBMS Show environment for cloudrdbmspackversion: '#{cloudrdbmspackversion}'"
+log "CloudRDBMS Show environment for playbook: '#{playbook}'"
+
+# if we do a deployment using multiple-clouds, this below will work, it will "see" all IP addresses
+string_of_ips = ""
+
+File.readlines("/app/listofHostsInCluster.log").each do |n|
```
sjangra-git: Question: If this code block for getting string of IPs from listofHosts... file is common in other files too, can we instead have this as a common ruby method in a library and invoke the method instead of multiple copies of the code? It might save effort when updating this method.

components/cookbooks/cloudrdbms/templates/default/restart.erb
```diff
@@ -0,0 +1,37 @@
+{
```
sjangra-git: +1 
for these templates! Nice work

components/cookbooks/cloudrdbms/recipes/restart.rb
BrandonBenge: Yes, we should. Ill create a bug for followup on all your responses here that we can drag out a little.

components/cookbooks/cloudrdbms/recipes/stop.rb
BrandonBenge: So I found that the workorder for many of the items we are looking to migrate do not actually pass the IPs. So in some of the recipes I use nslookup on the full hostname. The reason I dont just use hostnames in concord is because it doesnt recognize the dns names. Also, I dont want to expand on the oneops pack in the area of automated simplification because eventually Im hoping we move all of the code out of this. I can explain when I get back from vacation but Scott showed me the work orders we will get from managed services.

components/cookbooks/cloudrdbms/recipes/restart.rb
BrandonBenge: Agreed, should we expand on this though or should we start looking to move over to a fully java developed process in managed services? Ill open this up on the followup bug.

PR #1336, opened 2018-02-02
add custom image resource group metadata to compute service

dandagopi:


dandagopi:
Closing this PULL request to handle  using naming conversion in recipe logic instead of Metada

PR #1362, opened 2018-02-22
Changed DHCP config to allow dns to update

CodyLeeWhite:


sleepyorion:
@CodyLeeWhite : Just keep in mind that this change will effectively prevent compute with preference (static setting - rejecting settings from DHCP server) from setting its desired configuration because DHCP will update config on every new lease.

PR #1368, opened 2018-02-27
[STRSIDECAR-184] readme file documentation for service-mesh component

vsrp1010:


flatmap13:
The documentation looks good but we need to add an explanation on how to setup the tenant list across environments using a OneOps variable. Currently, the documentation suggests that the setup in the design is sufficient for all environments but that is not the case if the `environmentName` reference isn't parametrized per OneOps environment.

components/cookbooks/service-mesh/README.md
```diff
@@ -0,0 +1,37 @@
+## Description
+
+This component installs the service mesh client on the computes to integrate the application cluster nodes with Walmart's Service Registry portal.
+
+## Prerequisites 
+
+1. Component is available for all packs with linux-based OS. It doesn't work with windows-based platforms yet and is a future work in pipeline.
+2. Java should be installed on the computes with minimum version **1.8.0_111**
```
flatmap13: Just like with the windows support, we should mention that the plan is to provide a runnable artifact which contains its own JRE.

components/cookbooks/service-mesh/README.md
```diff
@@ -0,0 +1,37 @@
...
+  * **Tenants:** Provide 3 values separated by space in the textfield.
+    * First value is _application key for user application in service registry portal_
+    * Second value is _name of environment in service registry portal to which this compute/cluster corresponds to_
+    * Third value is _the local http address that service mesh process will use to communicate with local application_
+    * One example of such values is: **BOOKSTORE-DEMOAPP dev http://localhost:8080**
+
+4. Click on **Show Advanced Configuration** checkbox to reveal additional fields that may need to be updated. <br>
+<img src="images/doc03.png" alt="Advanced config" width="50%" height="50%"/> <br>
+Update these fields as following:
+  * **Service Mesh Version:** Make sure this fields value is 1.3 or higher (as recommended by SOA team)
```
flatmap13: Is there a way for users to simply specify "LATEST"?

components/cookbooks/service-mesh/README.md
```diff
@@ -0,0 +1,37 @@
...
+  * **Service Mesh Root Directory:** Directory where service mesh related resources will be placed on computes in the application environment. User can keep the same value or update if desired.
+  * **Additional JVM Parameters:** Optional field to pass any other JVM parameters as recommended by SOA team
+  * **Override config YAML:** Optional field to override the mesh config yaml file if needed. This is to perform quick config change tests and is not recommended/expected to be used for general purposes. If selected, it will display another textbox to populate the config yaml which SOA team can provide.
+5. Click on _Save_ button to save this component to the design.
+6. Before committing the design, user will need to ensure that ports 4140, 4141, 9990 are opened by modifying the **secgroup** component in the platform design. Refer to the **Inbound Rules** section in the image below to see how these 3 ports should be added as 3 entries along with any other already present port entries:<br><br>
```
flatmap13: Port 4140 should not be opened (i.e. exposed), it is exclusively serving egress traffic and should be reachable only by the local application.

flatmap13:
We should add a link to SR documentation so that users can learn how to register and manage their app.

components/cookbooks/service-mesh/README.md
vsrp1010: No. We could have utilized the cloud service mechanism but that would affect everyone, and each version may not be compatible with corresponding component recipes.

vsrp1010:
@flatmap13 Incorporated the review comments. Please have another look now. Thanks.

PR #1369, opened 2018-02-27
Fast image base skip

CodyLeeWhite:


components/cookbooks/compute/recipes/base.rb
```diff
@@ -108,8 +117,22 @@
     Chef::Log.info("Syncing #{cookbook_path} ...")
     cmd = node[:rsync_cmd].gsub("SOURCE",cookbook_path).gsub("DEST","~/shared/").gsub("IP",node[:ip])
     result = shell_out(cmd, :timeout => shell_timeout)
-    Chef::Log.info("#{cmd} returned: #{result.stdout}")
-    result.error!
+
+    # It is possible to get here and ssh not be ready. This guards against that.
```
dschelchkov: Is this guard still needed? You've added an extra wait for ssh to be ready in Line 25

PR #1404, opened 2018-03-15
new subscription design resourcegroup and availabilityset changes

dandagopi:


components/cookbooks/azure_base/libraries/utils.rb
```diff
@@ -136,17 +136,30 @@ def get_resource_tags(node)
...
     return tags
   end
 
+  def is_new_cloud(node)
+
+    cloud_name = node['workorder']['cloud']['ciName']
+
+    if cloud_name = ~/^azure-.*-wm-.*$/
+      is_new_cloud = true
+    end
```
bilal-naeem-confiz: This `end` should not be here. Move it to after the `else` block

PR #1414, opened 2018-03-16
Common NSGs for New Azure Subscription Design

maham-nazir-confiz:
For the new Azure subscription design, we are re-using network security groups across deployments in one region.
If deployments have similar NSG rules, they all get attached to one NSG that resides in a separate resource group for NSGs.
The NSGs are first filtered out on the basis of pack name and then rules are matched to identify the correct NSG.


dandagopi:
@maham-nazir-confiz please incorporate code review changes.

maham-nazir-confiz:
Changes have been made to use Platform CI-ID in NIC's name as per discussion.
@dandagopi please review the changes made.

PR #1417, opened 2018-03-17
azurekeypair deletion for new azure pattern and backword compatibility

shantanukande:


dandagopi:
@shantanukande please take care of my review comments

PR #1429, opened 2018-03-22
Refactored regex for fast image

CodyLeeWhite:


CodyLeeWhite:
cleaned up the kitchen test as well.

PR #1432, opened 2018-03-26
Changed the default os to be centos 7.3 instead of 7.2

CodyLeeWhite:


mtunganati:
Do we have 7.4 on all clouds ? This can only merged once we have teamplates of the clouds updated to have 7.4 image maps and pushed to prod

mtunganati:
These are the changes you need to do to push 7.4/7.3
1. make sure there is 7.3/7.4 on all cloud including AZ
2. Update the cloud templates to have 7.3/7.4 mapping
3. Make OS metadata changes to have 7.3/7.4 in the drop down 
4. Update the default to 7.3/7.4 in the base
5. Push all these changes together and update the cloud services

CodyLeeWhite:
1. √
2. already there
3. already there
4. √
5. all the changes are here. not sure what this entails exactly

mtunganati:
LGTM

PR #1438, opened 2018-03-28
Secgroup delete fix

bilal-naeem-confiz:
This PR contains the following
- If the Network Security Group does not exist, skip deletion and move on


dandagopi:
@bilal-naeem-confiz  please validate with new subscription design code for backward compatability

PR #1443, opened 2018-03-29
Customized Solr monitoring with JMX and Cluster status API for node status

minutheresathomas:
The solr monitor code would be outside the pack. This needs to be added as a jar to the solrcloud component.


spendap:
review+

PR #1446, opened 2018-04-02
change the new cloud regex format

dandagopi:


components/cookbooks/azure_base/libraries/utils.rb
```diff
@@ -152,7 +152,7 @@ def get_resource_tags(node)
 
   def is_new_cloud(node)
     cloud_name = node['workorder']['cloud']['ciName']
-    cloud_name =~ /^azure-.*-wm-.*$/ ? true : false
+    cloud_name =~ /^(dev|prod|stg)-az-(.*)-\d*$/ ? true : false
```
sureshg: `.*` includes space. Should we go for `(\S*)` - the non-whitespace char?

dandagopi: ^(dev|prod|stg)-az-(\S.*)-\d*$ is giving me better results, for example without . matching following string too (dev-az--1) @sureshg

PR #1469, opened 2018-04-17
Found Bug with forloop that needs addressed

BrandonBenge:


BrandonBenge:
Updated for bug https://jira.walmart.com/browse/STRCRDBMSD-3093 .

PR #1470, opened 2018-04-17
Fixed an issue with nodetool_repair platform procedure.

dschelchkov:
Its definition was still referencing walmartlabs bom class.
Changed to oneops.1


ajweave:
LGTM.  Thanks.

PR #1497, opened 2018-05-02
Fix Availability Set Issue in Storage Add Recipe

maham-nazir-confiz:
`storage::add.rb` recipe was using resource group's name as availability set's name to get the availability set for data disk's creation. 
However for the new Azure subscription design, resource group's name can be different from availability set's name depending on whether it is an old cloud or new cloud. Hence new code has been added to get availability set's name directly from `AvailabilitySetManager` class instead of using resource group's name.

KCI tests have been updated as well.


maham-nazir-confiz:
@dandagopi please review

PR #1524, opened 2018-05-16
script added to find segment size greater than 500MB

ladyjordan:


ladyjordan:
kafka_segment_size.sh added and broker.rb updated

PR #1531, opened 2018-05-22
Kitchen CI Test Case to Test Name Servers Added

maham-nazir-confiz:
Kitchen CI test case to check whether the name servers available in `/etc/resolv.conf` are reachable via port `53` has been added.


maham-nazir-confiz:
@gauravlall Please review

PR #1540, opened 2018-05-25
Added support for custom attributes for netscaler monitors.

dschelchkov:
Provides an ability to specify multiple custom attributes for each monitor,
by using port|attr_name => attr_value syntax in ecv_map attribute of lb component.


gauravlall:
@dschelchkov I believe we dont require this PR now

PR #1551, opened 2018-06-04
perform fqdn cleanup during compute delete

sleepyorion:


components/cookbooks/compute/recipes/delete.rb
```diff
@@ -30,3 +30,53 @@
 else
   include_recipe "compute::del_node_fog"
 end
+
+def get_records(record_type,ipaddress)
+  return node.infoblox_conn.request(:method => :get, :path => "/wapi/v1.1/record:#{record_type}?ipv4addr=#{ipaddress}")
+end
+
+def delete_record(record_ref)
+  res = node.infoblox_conn.request(:method => :delete, :path => "/wapi/v1.0/#{record_ref}")
```
sureshg: Why we need to use two WAPI version `1.0` and `1.1`. IMO, better to use the latest `2.5`

components/cookbooks/compute/recipes/delete.rb
```diff
@@ -30,3 +30,53 @@
...
+      when "a"
+        record_value = "ref => #{record['_ref']}, name => #{record['name']}"
+      end
+
+      Chef::Log.info("Found record :: #{record_value}")
+
+      res_delete = delete_record(record["_ref"])
+
+      Chef::Log.warn("Failed to remove record #{record["_ref"]}") if res_delete != 200
+      Chef::Log.info("Sucessfully remove record #{record["_ref"]}") if res_delete == 200
```
sureshg: Just warning required if it fails to delete an existing `ptr` or `a` records? Sorry, I don't remember what we discussed for this.

components/cookbooks/compute/recipes/delete.rb
```diff
@@ -30,3 +30,53 @@
...
+  res = node.infoblox_conn.request(:method => :delete, :path => "/wapi/v1.0/#{record_ref}")
+  return res[:status]
+end
+
+delete_vm_ip = node[:workorder][:rfcCi][:ciAttributes][:dns_record]
+cloud_name = node[:workorder][:cloud][:ciName]
+provider_service = node[:workorder][:services][:dns][cloud_name][:ciClassName].split(".").last.downcase
+
+# Support only infoblox cleanup
+unless !provider_service.eql?("infoblox")
```
sureshg: Instead of the double negative, I would use `if` 😃

PR #1571, opened 2018-06-15
updating the zone_info_missing as self, but in the caller added self class to call this method

minutheresathomas:


swathisingamsetty12:
review ++

PR #1576, opened 2018-06-20
Copying the certs to the compute

msumankumar:


components/cookbooks/secrets-client/recipes/set_attributes.rb
```diff
@@ -11,5 +11,5 @@
         Chef::Application.fatal!(msg)
 end
 node.set[:provider] = provider
-
+node.set[:persist_cert] == false
```
jayeshpise: @msumankumar shouldn't it be "=" instead of "=="

PR #1579, opened 2018-06-25
Servicemesh: use logrotate.

p4ali:
Change the default log location to /log partition. And set the default logrotate params.


vsrp1010:
The log rotate works, this PR can be merged.

PR #1580, opened 2018-06-25
Don't merge: creating test VIP for dcloadbalanver

shantanukande:


shantanukande:
Creating backup Test VIP
-------------------------
1. Create backup test VIP at dc level
- have IP as nil
- port as 0
- listenpolicy as 'backup_delete_lb_group'
(having set this as listen policy will restrict traffic from GSLB to this test VIP)

2. Bind all service group under that DC to test VIP irrespective of primary and secondary cloud
3. In primary VIP add corrospoding test VIP as 'backupvserver'.
4. Making cloud primary or secondary should not have any effect on test VIP.
5. Same goes for binding/unbinding clouds

For simplicity, we are not doing any changes in UI for user to select option to create Backup VIP, we are doing it on basis of flag from CMS.
Backword compatibility: User can touch LB component and backup VIP will be created and added as backupvserver to primary VIP.

6. Kitchen-ci test cases

PR #1581, opened 2018-06-25
Minor Code fix when total no of ips across fault domains are unequal

swathisingamsetty12:
With this fix the deployment will complete but we should select replicas manually for shards which have replicas less than replicator factor.


swathisingamsetty12:
This repo is no longer accessible to us. Closing this PR

PR #1600, opened 2018-08-06
Add FairScheduler configs

weiatwork:
By default Yarn is using capacity scheduler, and without configuration, any preceding job will take all the resources of the cluster and other jobs will just wait.

Better use fair scheduler as that one works better.


weiatwork:
Btw, this is related to DTBFDP-615

PR #1601, opened 2018-08-09
Updated the Kafka pack with Kafka 1.1.1 version

rbaddam:
Updated the Kafka pack with Kafka 1.1.1 version and also changed the Binaries location for Kafka versions from 1.0.0


mosabua:
Adding the beta version when the final version is available is probably a bad idea.

PR #1606, opened 2018-10-23
Enterprise Server-Monitor script work with HTTPS on the local instance

bbourqu:
* Works for HTTPS, HTTP, HTTPS/HTTP will choose HTTPS to monitor when
  both are chosen
* Remove the runOnEnv to support internal processes


bbourqu:
Closing due to this repo being located elsewhere

